PR #12, opened 2018-10-11
Add clarifications for links

sypets:


sypets:
I would like to merge this asap. Someone asked me about a core patch in the changelog about links and the docs where not clear enough. 

@marble I am not sure: what do we recommend for external hyperlinks? You can add one or two `_`. I explained the difference. I usually use 2. If someone does not know the difference and does not want to dive into the intricacies, I think it would be good if we can just say: if in doubt, do it like this.

marble:
This will ALWAYS work:
backtick, linktext, blank or newline, <, link target (url), >, backtick, underscore, underscore
I would recommend this as the "will always work" solution. That's what you want to know as a beginner. Everything else is just optimising. Worst that can happen is unessaray source.

PR #18, opened 2018-11-24
Fix syntax in sheme for link

jokumer:


sypets:
Yikes! Thanks for submitting this!

PR #29, opened 2018-11-24
Fix external link for 302

jokumer:
External link has been changed and redirects with 302


sypets:
I am making considerable changes in the WritingDocsOfficial section. This change will conflict with my changes #31. I am not quite done yet and might push more additions to #31, but would appreciate it if you could take a look, when it's done. Thanks. 

Will close this PR.

PR #66, opened 2019-05-16
Add information about reviewing manuals

sypets:
Related: https://github.com/TYPO3-Documentation/T3DocTeam/issues/73


DanielSiepmann:
I like the approach, it's the same as on Wikipedia, or at least it's similar.

Documentation/GeneralConventions/ReviewInformation.rst
```diff
@@ -0,0 +1,73 @@
...
+Here are some examples:
+
+
+Add Review Information to a Page
+================================
+
+Example for Good Content
+------------------------
+
+.. note::
```
DanielSiepmann: I'm not sure about the approach to add "correct" to pages. This will just clutter the content and is only valid until the next change is introduced (That's why the concrete version and date is needed to be added).

Also this might give the feeling everything else is not correct.

I would only add the warnings I guess.

sypets: Yes, that is a good point. My motivation was to try to find an approach for systematically reviewing an entire manual which is currently not entirely up-to-date for a branch (like "Extbase / Fluid book" or "TYPO3 Explained"). If one person reviews entire manual until it is correct, we do not need this. If there are several people doing this, we need some method of tracking what is already been reviewed. Otherwise we have some pages being read again and again while others are sadly lacking. But I suppose the coordination could also be done elsewhere. I will push another change.

DanielSiepmann: I understand. We could use the `todo` directive. This can be turned on and off during rendering. So one could, maybe, add todo annotations to all pages. Once a section is checked, this can be removed. This way one can see what sections still need to be checked. For local rendering this can be turned on, while kept off on production.

sypets:
I am currently not sure how to handle the reviewing of manuals. As of now, we neither have a workflow for this, nor people doing this. I no longer think it is a good idea to add the review information to the page. 

I will reduce this PR to information that is already pretty much agreed on and remove the rest.

PR #87, opened 2019-06-03
fix typo

math-uzh:


sypets:
whoops. We must have missed this. Thanks for fixing! :thumbsup:

PR #89, opened 2019-06-06
Remove misleading text

sypets:


DanielSiepmann:
Thanks for removing this one. I didn't understand why it was added at all :+1:

PR #98, opened 2019-06-16
Add documentation for webhook in GitLab

DanielSiepmann:


sypets:
@DanielSiepmann 

>  in XXXXXXX steps:

Was that intentional? Otherwise, I would just remove it.

DanielSiepmann:
I've fixed that, it was a placeholder to contain the number of steps.

Issue #42, opened 2018-12-30
How to write & markup menu items


sypets:
Sometimes you want to point out the steps of opening a menu and you need the complete menu path. How should this be written?

sypets:
Suggestion: 

* Separate path segments with **>** 
* write entire phrase in italics.
* Use spelling as used in the GUI

 _File > Open > New_

sypets:
see https://docs.typo3.org/typo3cms/HowToDocument/GeneralConventions/ContentStyleGuide.html#menu-and-clickpath-style

PR #2, opened 2018-07-03
[TASK] Link the whole "See" label

mbrodala:

PR #102, opened 2019-06-21
Add alternative method for documentation rendering

codemonkey1988:
On some Macs, the source command will fail. Therefore an alternative method is provided.

### DIFF
--- a/Documentation/RenderingDocs/Quickstart.rst
+++ b/Documentation/RenderingDocs/Quickstart.rst
@@ -33,6 +33,10 @@ Commands to Render Documentation
 
    docker pull t3docs/render-documentation
    source <(docker run --rm t3docs/render-documentation show-shell-commands)
+   # If the command above will fail, please run the following three commands:
+   # mkdir -p ~/bin
+   # docker run --rm t3docs/render-documentation show-shell-commands > ~/bin/t3docs
+   # source ~/bin/t3docs
    dockrun_t3rd makehtml
    xdg-open "Documentation-GENERATED-temp/Result/project/0.0.0/Index.html"
 
--- a/Documentation/RenderingDocs/Quickstart.rst
+++ b/Documentation/RenderingDocs/Quickstart.rst
@@ -29,16 +29,22 @@ Commands to Render Documentation
 ================================
 
 .. code-block:: bash
-   :linenos:
 
    docker pull t3docs/render-documentation
    source <(docker run --rm t3docs/render-documentation show-shell-commands)
-   # If the command above will fail, please run the following three commands:
-   # mkdir -p ~/bin
-   # docker run --rm t3docs/render-documentation show-shell-commands > ~/bin/t3docs
-   # source ~/bin/t3docs
    dockrun_t3rd makehtml
    xdg-open "Documentation-GENERATED-temp/Result/project/0.0.0/Index.html"
+
+.. note::
+
+   If above command does not work, use the following instead:
+
+   .. code-block:: bash
+
+      mkdir -p ~/bin
+      docker run --rm t3docs/render-documentation show-shell-commands > ~/bin/t3docs
+      source ~/bin/t3docs
+      xdg-open "Documentation-GENERATED-temp/Result/project/0.0.0/Index.html"
 
 Explanations, the numbers correspond to the line numbers in code snippet:
 


DanielSiepmann:
Thanks for adding this alternative method. That should help a lot of people.

@marble would you take a look why this is necessary nowadays? I guess it wasn't necessary with version `v1.6.11-html`, but I'm not sure.

sypets:
We have had reports from several people that the source command did not work for them on Windows / Mac platforms. That was also the case with previous versions.

See also this (closed) issue with very similar solution: https://github.com/t3docs/docker-render-documentation/issues/2

As alternative we also tested with Docker compose. That is also documented in the READMe, I think. I do think what Tim suggested above is a good solution and similar to the solutions that had already been recommended. (I work on Linux and didn't have these problems and hadn't been aware of them when I originally wrote the text).

See also existing issues (some of them closed):

problems:

* macOS: https://github.com/t3docs/docker-render-documentation/issues/2
* broken pipe on macOS: https://github.com/t3docs/docker-render-documentation/issues/42
* problem with Gitbash / Windows 10: https://github.com/t3docs/docker-render-documentation/issues/16

alternatives:

* docker compose: https://github.com/t3docs/docker-render-documentation/issues/14
* docker compose: https://github.com/t3docs/docker-render-documentation/issues/26
* develop starter script for windows: https://github.com/t3docs/docker-render-documentation/issues/13

See:

*

n3amil:
Didn't we had this fixed already? I would suggest to do less as possible on docker host, like "source <(docker run --rm t3docs/render-documentation:v2.1.0 show-shell-commands)"  but straight up within docker. Thats what from my point of view docker is for and now we got the opposite again. I would think with docker we really want to get away from "works on my machine". Sure performance and file permissions is also a thing, which is not exactly the same on linux/mac/windows but this is not a problem for docs rendering i think.

sypets:
@n3amil 

I understand your irritation. Some things have been switched back and forth. I think the reason is we have not found an optimal solution yet (or not everyone agreed on the same optimal solution).

It would probably be best to leave the documentation alone for now and **first** decide on the optimal solution. I will take this up with the Documentation Team and we will see what we can do.

If I understand you correctly, you would suggest to use docker compose instead of the "source" command? Is that correct?


> I would think with docker we really want to get away from "works on my machine"

Yes, theoretically, that is a great idea. Unfortunately, in practice, it doesn't always work. We have Docker, which takes away the problem of installing packages with specific versions on already existing systems. But as has been the case here, other problems appear. 

> If i can just run docker-compose run makehtml and it will work for everyone - why dont use it?

The question is, what is better: using docker-compose or the (originally used) source command. 

I think with docker compose, we have not had the chance to test this a lot. It basically seems like a good idea to me. I don't know if this will work reliably on all platforms (Linux, Mac, Windows).

It worked for me and I would agree it is a good solution. But, very few people gave feedback or commented on this so far.

I work on Linux. I don't have other platforms available.

> But why do we have a "Writing Documentation ➜" and "Documentation Rendering With Docker ➜" next to each other in the tutorials section?

I originally thought we would use the [GitHub README](https://github.com/t3docs/docker-render-documentation/blob/master/README.rst) as primary source for rendering. Main reason being: put the docs where the source is. But in practice I found it not ideal, because:

1. everything about TYPO3 doc is in "Writing Documentation" and then you redirect people to a different Readme on GitHub with totally different rendering. 
2. we already have various information in several places. It is confusing. You have to say, ok for this you go to "Writing Documentation" and then rendering is documented in the README. I would recommend to give up the README in favor of "Writing Documentation", but I can't decide this myself. 

Please have a little patience. We will hopefully sort this out.

n3amil:
> Yes, theoretically, that is a great idea. Unfortunately, in practice, it doesn't always work
This is because for example with the source command we expect certain stuff to work on every platform which is not the case obviously.
If you have docker/docker-compose installed and we suggest to run docker commands to render the docs it will work on every platform because its a specific command for docker/docker-compose and not something depending on the platform itself. 
I Am trying to point out my thoughts and how we can improve this docker rendering stuff. In my opinion, which i already pointed out in the linked issues, creating docs should be as easy as possible. If we can have a 3 liner for rendering docs on linux, mac and windows i would prefer it this way.

DanielSiepmann:
I would go for @n3amil approach and replace our examples with the single solution, that works on all machines. As for the named reasons, one solution that works all the time, and is easier to maintain.

I've never used docker compose until now.

Placing https://github.com/t3docs/docker-render-documentation/issues/14#issue-347632042 in the manual root and running `docker-composer up` worked for me out of the box on Ubuntu.

sypets:
@DanielSiepmann The single solution would be docker-compose, right? 

This means we suddenly replace the commands that people have been using all along.

I agree, that it is probably the best solution long term, but I would not suddenly replace one solution with another that has not been tested very much yet. 

I would like to add the 2 solutions for now, side by side. Each one on a separate page. Allow for some more testing and eventually remove the source solution.

I know having 2 solutions is not so ideal: I am always saying we should simplify things. But better than confusing people by suddenly switching to something still unfamiliar.

I already started a PR for this. I will post that next week. 

Just to make sure, we are talking about the same thing:

1. Solution 1 (has been used previously and causes problems on some machines, esp. Window)

```
source <(docker run --rm t3docs/render-documentation show-shell-commands)
dockrun_t3rd makehtml
```

2. Solution 2 (should work on all machines but has not been tested yet by a lot of people)

with docker-compose.

One problem, I did have with docker-compose: When I use that, the files generated in my Documentation-GENERATED belong to root (on Linux). If I subsequently want to remove them (as plain user), I can't (well not without sudo). Removal is sometimes necessary, if a menu changes or a reference. It's not a huge problem, but it's ugly and it would be great if we can find a solution for that, too. I suspect, it should be possible to make that work with better configuration.

```
$ ls -l Documentation-GENERATED-temp/Result/project/0.0.0/

drwxr-xr-x  2 root root   4096 Feb  2 10:19 Account
drwxr-xr-x  2 root root   4096 Feb  2 10:19 AddingDocumentation
drwxr-xr-x  8 root root   4096 Feb  2 10:19 Appendix
drwxr-xr-x  2 root root   4096 Feb  2 10:19 BugfixingAZ

```
I also used https://github.com/t3docs/docker-render-documentation/issues/14#issue-347632042

To further test the methods and get feedback there is already an issue: [Test rendering workflow with "docker-compose" vs "source + docker run" #26](https://github.com/t3docs/docker-render-documentation/issues/26). I would link to that on the documentation pages.

Please wait for my PR.

Thank you @n3amil for pointing this out and being persistent!

sypets:
@n3amil question:

`docker-compose run --rm <service>
`
`docker-compose up`

Is that basically the same?

sypets:
I added the PR #108

n3amil:
> @n3amil question:
> 
> `docker-compose run --rm <service> `
> `docker-compose up`
> 
> Is that basically the same?

docker-compose run --rm creates the container, executes the build and removes the container afterwards. This is the correct way to do it. 
For the permission issue - docker-compose run --rm --u 1000 <service>  should do the trick. 1000 should be the uid/name of your current user. This is not needed for Windows (on Docker for Windows you dont care about file permissions), i dont know about MacOS.

sypets:
Hmm, a little more tricky: Using the -u option results in errors in the rendering. I will do some troubleshooting to find out more. If you can come up with a solution, that would be great Additional tip: Creating a `tmp-GENERATED-temp` will cause this directory to not be deleted after the rendering. There you can find additional logs (but I think in this case it is probably not necessary). See https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/Troubleshooting.html#get-detailed-information

```
$ id --user
1000
```

`$ rm -rf Documentation-GENERATED-temp/`

(for testing it is important to remove Documentation-GENERATED-temp. The image does not always rebuild everything from scratch, it caches some things (which is good). So to get a full rebuild, it is necessary to remove the generated files first)


`$ docker-compose run --rm -u 1000 t3docmake`

```
==================================================
   14-Provide-the-project/40-Steps-B/run_32-Check-rebuild-needed.py
Traceback (most recent call last):
  File "/ALL/Toolchains/RenderDocumentation/14-Provide-the-project/40-Steps-B/run_32-Check-rebuild-needed.py", line 151, in <module>
    with file(checksum_file, 'wb') as f2:
IOError: [Errno 13] Permission denied: u'/ALL/Makedir/build.checksum'
   exitcode: 1

...

==================================================
   18-Make-and-build/40-Html/run_40-Make-html.py
Traceback (most recent call last):
  File "/ALL/Toolchains/RenderDocumentation/18-Make-and-build/40-Html/run_40-Make-html.py", line 149, in <module>
    os.makedirs(outdir_in_cache)
  File "/usr/local/lib/python2.7/os.py", line 150, in makedirs
    makedirs(head, mode)
  File "/usr/local/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/RESULT/Cache'
   exitcode: 1

...


We saw these exitcodes (code, count):
{
  "0": 57,
  "1": 3,
  "22": 6
}
exiting with exitcode 0

Final exit status: 0 (completed)

```

n3amil:
I just tested this on linux and i got the same behavior. What i don't understand - the docker run command needs to do address this issue too. This permission issue on linux is something i already encountered multiple times and as far i know there is no simple solution for that. But with the docker run command there is already a solution in place and it should be possible to use just that.

PR #121, opened 2019-08-21
Add chapter for updating docs

sypets:
Additional notes:

* about deprecation topic (e.g. remove deprecated information vs.  provide information about deprecations and upgrade hints), we could link to some examples
  * here, good (helpful) example: https://docs.typo3.org/m/typo3/reference-typoscript/9.5/en-us/Conditions/Index.html


Documentation/HowToUpdateDocs/Index.rst
```diff
@@ -0,0 +1,25 @@
...
+===========================
+
+Once a new TYPO3 release comes out, the main documentation (e.g. :ref:`t3coreapi:start`,
+:ref:`t3tca:start` etc.) must be updated.
+
+Here, we describe some best practices for updating the official documentation
+for a new TYPO3 release. We stick to the core conventions as much as possible because that
+makes it easier for everyone.
+
+* Once a feature is deprecated in the core, it is removed from the main documentation
```
susannemoog: I'm unsure about that one. For user experience sake I would prefer to document the deprecation _and_ the alternative to make migration easier - as I'm guessing the docs are used by new and upgrade users.

sypets: We already talked about this during the Slack call. The result was that we roughly agree with you. 

I think the main reason to remove the deprecated functionality in the past was that this way the docs do not have to be updated twice (once for deprecation and then for removal). 

I think you can find pros and cons for both arguments. 

But, I agree that having information about deprecated functionality (marked as deprecated) is more user-friendly. I So I would agree to do it this way and if there are still objections we could hash it out on decisions.typo3.org.

Here is the link to the issue about versioning that was also brought up: https://github.com/TYPO3-Documentation/T3DocTeam/issues/14

sypets: I am currently wondering: Is "Writing Documentation" really the best place for the text proposed with this PR? Or should we move it to general information about creating a new TYPO3 release? I assume this already exists. I think it would be great to have all information about creating a release in a central place, including (or linking to) information about Creating what's new slides, branching out the documentation, updating documentation etc. I think this could be considered a part of creating a new TYPO3 release and documentation would be a subtask. 

@susannemoog What do you think? Should we proceed here or somewhere else?

susannemoog: I'm not sure that there is any public information about creating a release, as that is usually done team internal. So in general I'm fine with adding it here. 

On the deprecation topic: let's continue in the other issue.

sypets: I changed the text to recommend documenting the deprecations. Also, I linked to some of the still open issues with further suggestions, so we continue optimizing this at a later time.

PR #124, opened 2019-08-27
Minor language improvements

sypets:
Some fixes, changes for clarity and removal of superfluous text

- shorten text
- add comma
- replace word
- move new important block down so that section about the hosters is not split up. 
- replace for "in those systems" for more clarity
- remove "An update is required". Sentence seemed a little vague and not really clear (as if another step is required) and not necessary. All people need to know: legacy webhook -> no longer supported, use new webhook.
- Add extra text (For the documentation to be rendered ...) to creation of GitHub hook as well (because it already existed in the description for bitbucket and gitlab).

### DIFF
--- a/Documentation/WritingDocForExtension/Webhook.rst
+++ b/Documentation/WritingDocForExtension/Webhook.rst
@@ -36,8 +36,8 @@ the necessary hook for the respective hoster.
 
 .. _webhook-how-webhook-works:
 
-How webhooks works
-==================
+How Webhooks Work
+=================
 
 If it's your first time working with webhooks, take a look at `GitHub developer
 <https://developer.github.com/webhooks/>`_. The configuration below triggers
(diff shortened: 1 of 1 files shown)


Documentation/WritingDocForExtension/Webhook.rst
```diff
@@ -15,34 +15,33 @@ following hosters:
...
    <https://my.typo3.org/index.php?id=35>`__.
 
-The following chapters provide illustrated walkthrough tutorials on how to add
-the necessary hook in those systems.
-
 .. _webhook-how-webhook-works:
 
-How webhook works
-=================
+How webhooks works
```
susannemoog: if webhooks is plural then "works" needs to be "work"

sypets: right. I'll fix it.

PR #127, opened 2019-09-12
[TASK] Add missing type attention

gilbertsoft:


sypets:
Actually, there are some more admonitions not listed on that page. We can list all of them, but the syntax is basically always the same. 

I would just list the ones we most commonly use and possibly point out, you can use all admonitions which are supported (as listed in http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions).

I will add the link to your PR.

I personally would not use all admonitions because some basically mean more or less the same and I find the docs more easily readable with just a small subset of markup. E.g. attention, caution and danger would probably be interchageable.

PR #128, opened 2019-09-28
Solve an issue using keywords

linawolf:
not everyone writing documentation is using github every day

### DIFF
--- a/Documentation/GitHub/Index.rst
+++ b/Documentation/GitHub/Index.rst
@@ -35,8 +35,8 @@ and click on :guilabel:`"Repository"`.
 
 .. _links-github-issues:
 
-Issues
-======
+Find Issues
+===========
 
 Pick an issue for a topic you are familiar with and try to fix it. Some of the
 issues address a problem, some are enhancements where new text needs to be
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
This is already documented at https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CommitMessages.html

Please do not add this at another place. Instead add a reference to the existing location. And add your additional information there.

Documentation/GitHub/Index.rst
```diff
@@ -161,3 +161,8 @@ About reviewing pull requests, you can look in the GitHub help pages:
    If you are reviewing a pull request and want to leave comments, make sure to
    `mark the line <https://help.github.com/en/articles/commenting-on-a-pull-request>`__
    in the PR.
+   
+Solve an Issue
+==============
+
+When you solve an issue from the github simply write solves #[issue id] into the commit message. See `Closing issues using keywords <https://help.github.com/en/articles/closing-issues-using-keywords>`__.
```
sypets: By convention, we use `Resolves: <issue number>`. This has the advantage that it is the same syntax as for core development. GitHub will automatically close the issue, when you use this (I think I tested it).

This is also documented here: https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/CommitMessages.html

Can you update your change?

PR #129, opened 2019-10-07
Adjust information about custom infrastructure for webhook

DanielSiepmann:


Documentation/WritingDocForExtension/Webhook.rst
```diff
@@ -8,16 +8,16 @@ Webhook
...
   - :ref:`webhook-github`
   - :ref:`webhook-bitbucket-cloud` and Bitbucket self-hosted
   - :ref:`GitLab Cloud <webhook-gitlab>` and :ref:`GitLab self-hosted <webhook-gitlab>`
 
-If none of the above is used, a mirror to one of these is recommended. Also the
-`Documentation Team <https://typo3.org/community/teams/documentation/>`__ can
-setup custom hostings.
+If none of the above is used, a mirror to one of these is recommended. This is
+to ensure people know where to find extensions. Also maintenance of further APIs
+would be to expensive.
```
sypets: > to expensive 
too expensive

> If none of the above is used, a mirror to one of these is recommended.

Actually, it is not just recommended. It is necessary in order to render docs.

> If none of the above is used, a mirror to one of these is recommended. This is to ensure people know where to find extensions. Also maintenance of further APIs would be to expensive.

This sounds a little like unnecessary justification. If we currently only support the 3 hosters and nothing else I would just state it as a matter of fact. I would just write:

If none of the above is used, you must setup a mirror using one of the supported hosters in order to enable  documentation rendering on docs.typo3.org.

PR #149, opened 2020-03-26
[FEATURE] Add script to trigger Intercept webhook

garbast:


sypets:
The [Troubleshooting](https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/RenderingDocs/Troubleshooting.html) page is for local rendering (as is the entire chapter). 

If that is ok for you, I would just rearrange your text a bit. I would move it to the [Extension documentation chapter.](https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingDocForExtension/Index.html) That is also where the info for the webhook etc. is.

garbast:
That's fine to me. I was ask if i could prepare something and this is was i thought that it could match. But you are right the extension chapter would match better.

DanielSiepmann:
Thanks for providing the script and PR :+1: I guess that will help some more extension authors and reduce questions and discussions on our end.

PR #155, opened 2020-04-30
Fix typo in trigger_documentation_push.sh example

bnf:
Also document invocation via `sh`.
An alternative would be to document
`chmod +x trigger_documentation_push.sh`,
but `sh` prefix is more straight forward.


DanielSiepmann:
Thanks for the fix and improvement.

PR #157, opened 2020-05-20
Remove unneded indentation

tmotyl:
on https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/DefinitionLists.html#example-2-nicely-styled

some code examples are styled as quotes


tmotyl:
I'm not sure it's the correct way to fix it, so please review.

PR #160, opened 2020-07-20
Add page for backporting changes

sypets:


Documentation/Maintainers/BackportChanges.rst
```diff
@@ -0,0 +1,72 @@
...
+.. code-block::
+
+   Author: Author <email>
+   Date:   Fri Jun 26 15:50:08 2020 +0200
+
+      [TASK] Subject
+
+      (cherry picked from commit 609493dd8893cbac7ad78aa38a23e02d011bb0c2)
+
+**How do we keep track of what is already backported?**
```
susannemoog, line 62: As a side note: if the original PR is merged with a reference to the PR # in the commit, the cherry-picks will also be automatically be displayed in the PR which makes it even easier to see (for example https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-TSconfig/pull/128 )

susannemoog (approved):

Documentation/Maintainers/BackportChanges.rst
sypets: That looks great. How did you do that? (It looks like the commit message must be changed or is there an automatism that does that?)

I am ok if we change my original suggestions.

susannemoog: When merging I usually chose "Squash and Merge" which squashes changes down to one commit and adds the original text and PR number to the commit headline. 
I then cherry-pick (without anything special) that one commit which gets exactly the same commit message with the issue reference -> pushing that adds that to the orginal PR.

sypets: Ah, cool. I often use "Squash and merge" when there is more than one commit. I didn't notice that it adds the PR number to the commit headline.

I like this solution better than mine, will change my PR later.

sypets: This also means the label "backporting done" which I have partially used is not necessary. I will remove that.

mbrodala: If it is useful you can make "Squash and Merge" the default, see:

https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

Issue #164, opened 2020-08-21
Add PlantUML diagrams docs


alexander-nitsche:
Add docs for diagrams which are rendered by PlantUML via the Sphinx extension sphinxcontribution-plantuml . This feature is required to inform documentation authors about the availability of diagrams for their reST documentation. The diagram feature is provided by t3docs/docker-render-documentation#97.

PR #165, opened 2020-08-21
Add PlantUML diagrams docs

alexander-nitsche:
image

Resolves: #164
Releases: master


alexander-nitsche:
Daniel Siepmann:

> Looks like the sections are sorted from easy to complex. I would add the section below "Special Character".

alexander-nitsche:
Daniel Siepmann:

> And maybe add an info to that section on `Codeblocks with Syntax Highlighting` as I think it is related to each other. 
> The idea is: UML is tightly coupled to code documentation, like program flow, or structure like class diagrams. So people looking for how to document code, might be interested that they can document via UML as well. Therefore a short link to the uml section might be helpful

susannemoog:
This is very very cool, I love it! Just tested it by building the docker container from develop myself ;) I guess we need to wait with merging until @marble releases the new container, but I'm already excited :-D

marble (commented):
Let's revisit this PR when the the new Docker container featuring Plantuml is released.

marble (changes requested):
① Please don't use the include directive for these examples. This has already been changed in the DemoDocs. https://github.com/TYPO3-Documentation/t3SphinxThemeRtdDemoDocs/blob/master/Documentation/Directives/sphinxcontrib-plantuml.rst
② I have the plan to split the demodocs into two repos: One for testing purpose, the other for demonstration and user inspiration. The repo for testing should containe one example of each directive ony to be fast. The other one may render slowly but should contain all kind of insprigin examples. We could fille a complete book with plantuml und graphviz examples. Personally I doubt they should be part of the "How to document" guide.
③ On the other hand, "Inspiring examples of reST usage" could be a chapter of its own in this guide. I think, we should figure out in a conference call what we want to considere the best solution.

alexander-nitsche:
Ad 1) Will adapt the code.
Ad 2+3) I can see two purposes: Getting inspiration by as many examples as possible and testing diagram rendering for new docker image releases. For the first i would tend to replace the diagrams by real world diagrams every year by having a contest of the nicest new diagrams of the TYPO3 world. And for the latter i have no idea yet how this would be accomplished best: The last time i added a patch to the docker image i just compared manually all examples one by one.

alexander-nitsche:
Ad 1) Diagrams are now included inline in the ReST code.

marble (approved):
Great work, Alex!

Issue #176, opened 2020-10-28
Fix guideline of list indentations


alexander-nitsche:
Currently the documentation guide mainly requires sub lists to be separated by a blank line and indented by 3 white spaces. This is true for the blank line but not always for the indentation: The sub list should always be aligned with the first letter of the text of the parent list item.

For example:

```
* Parent list item text

  * Sub list item indented by 2 ws

*. Parent list item text with a dot

   *. Sub list item indented by 3 ws
```

This issue is about reworking all lists and how-tos of lists in the guideline accordingly.

See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#bullet-lists .

marble:
I don't know what we are currently, writing in our guide, but I just want to throw in, that I, personally in my own documents, tend **to always use and indentation of three for lists.** 
This is because in .editorconfig we have that setting of three, and I can just type "*", TAB, Text of list item. This make rearrangement of multiline list items easier, as I can use TAB for this:

    1234567890

    *  The T at the beginning of this line is in column three, as there are "*", " ", " " precedding it
       The T at the beginning of this line is in column three.
       The T at the beginning of this line is in column three

.

alexander-nitsche:
Ok, makes sense. The above example would change then to
```
*  Parent list item text

   *  Sub list item indented by 3 ws

*. Parent list item text with a dot

   *. Sub list item indented by 3 ws
```
This would still require a rework of the lists but a different explanation in the how-tos of lists.

marble:
I fixed my example.

PR #177, opened 2020-10-28
Fix guideline of list indentations

alexander-nitsche:
Resolves: #176
Releases: master

### DIFF
--- a/Documentation/About.rst
+++ b/Documentation/About.rst
@@ -33,10 +33,10 @@ it cover-to-cover.
 
    Specifically, you might like to look at:
 
-   * :ref:`basic-principles` to familiarize yourself with the structure
-   * :ref:`conventions` for tips on coding guidelines, spelling etc.
-   * :ref:`reStructuredText & Sphinx Introduction <writing-rest-introduction>`.
-   * Additionally, keep the :ref:`rest-cheat-sheet` handy to look up the syntax.
+   *  :ref:`basic-principles` to familiarize yourself with the structure
+   *  :ref:`conventions` for tips on coding guidelines, spelling etc.
+   *  :ref:`reStructuredText & Sphinx Introduction <writing-rest-introduction>`.
+   *  Additionally, keep the :ref:`rest-cheat-sheet` handy to look up the syntax.
 
 
 But often, you would like to **jump right in**, start with a specific task
@@ -44,16 +44,16 @@ and learn as you go along. In that case, find your task here and start reading:
 
 
 
-* :ref:`writing-doc-for-ext-from-scratch` if you have an extension and would
-  like to write documentation for it using the sample extension manual.
-* :ref:`docs-contribute-github-method` if you would like to contribute to
-  the official documentation and need an easy introduction to editing
-  documentation directly on GitHub. This does not require any development
-  tools, you just need a browser and a GitHub account.
-* :ref:`docs-contribute-git-docker` if you would like to contribute to
-  the official documentation and are already familiar with Git and Docker
-* :ref:`render-documenation-with-docker` if you would like to start with rendering the
-  documentation locally with Docker.
+*  :ref:`writing-doc-for-ext-from-scratch` if you have an extension and would
+   like to write documentation for it using the sample extension manual.
+*  :ref:`docs-contribute-github-method` if you would like to contribute to
+   the official documentation and need an easy introduction to editing
+   documentation directly on GitHub. This does not require any development
+   tools, you just need a browser and a GitHub account.
+*  :ref:`docs-contribute-git-docker` if you would like to contribute to
+   the official documentation and are already familiar with Git and Docker
+*  :ref:`render-documenation-with-docker` if you would like to start with rendering the
+   documentation locally with Docker.
 
 .. _credits:
 
--- a/Documentation/Appendix/ResourcesForEditors.rst
+++ b/Documentation/Appendix/ResourcesForEditors.rst
@@ -9,8 +9,8 @@ Here, we list some resources that are useful when editing this manual:
 Resources for This Guide
 ========================
 
-* :ref:`Linktargets <linktargets>`  (labels for cross-referencing)
-* `warnings.txt <https://docs.typo3.org/typo3cms/HowToDocument/_buildinfo/warnings.txt>`__
-  (list of warnings)
-* `Issues on GitHub <https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/issues>`__
+*  :ref:`Linktargets <linktargets>`  (labels for cross-referencing)
+*  `warnings.txt <https://docs.typo3.org/typo3cms/HowToDocument/_buildinfo/warnings.txt>`__
+   (list of warnings)
+*  `Issues on GitHub <https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/issues>`__
 
--- a/Documentation/BasicPrinciples.rst
+++ b/Documentation/BasicPrinciples.rst
@@ -14,8 +14,8 @@ Introduction to TYPO3 Documentation
 For an excellent introduction, please see the 3 part series on Marcus
 Schwemer's blog TYPO3 Worx:
 
-* `GUIDE TO TYPO3 DOCUMENTATION – PART 1: THE USER-INTERFACE <https://typo3worx.eu/2017/01/guide-to-typo3-documentation-part-1-the-ui/>`__
-* `GUIDE TO TYPO3 DOCUMENTATION – PART 2: CONTENT STRUCTURE <https://typo3worx.eu/2017/01/guide-to-typo3-documentation-content-structure/>`__
+*  `GUIDE TO TYPO3 DOCUMENTATION – PART 1: THE USER-INTERFACE <https://typo3worx.eu/2017/01/guide-to-typo3-documentation-part-1-the-ui/>`__
+*  `GUIDE TO TYPO3 DOCUMENTATION – PART 2: CONTENT STRUCTURE <https://typo3worx.eu/2017/01/guide-to-typo3-documentation-content-structure/>`__
 
 .. important::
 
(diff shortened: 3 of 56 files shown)


alexander-nitsche:
I compared all changed pages with the previous state.

sypets:
Why the extra space?

alexander-nitsche:
> Why the extra space?

This is according to the issue [description by Martin](https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/issues/176#issuecomment-717780141): Sub list items have to be indented to the level of the parent list item **text**. If the parent list item text gets indented by 2 - including the list item sign - then the sub list items will have to be indented by 2 too. If the parent list item text gets indented by 3, e.g. if the list item sign has an additional dot, then the sub list has to be indented by 3 too. This variance is error prone and produced some bugs throughout the documentation. If the parent list item text gets always indented by 3, then the sub list can also always be indented by 3 and this can be achieved easily and reliably by pressing TAB after writing the list item sign (i am sure someone will need less words to describe it more precisely:).

sypets:
Will people use this type of indentation everywhere? Is this helpful? What do typical contributors say?

The sublist (level two and above) are a problem if not used correctly, but they are much rarer than the lists themselves. Should we change the indenting for lists everywhere - meaning not just in this guide but throuought the docs? 

I would make these kinds of changes when they are tools to enforce this and ideally makes the change automatically. Otherwise it just introduces a new way to do things, but you still have the "old" way in various places.

alexander-nitsche:
I think this TAB rule is quite easy to use in your daily documentation workflow. It is helpful to overcome inconsistencies in indentation which is not only used for sub lists. The alternative would be to indent with typing in multiple white spaces in order to align child items with their parents which has been obviously done until now and produced some formatting issues throughout the documentation.

But even if you would prefer to type in multiple white spaces for child items, this documentation mostly spoke about indentation of 3 white spaces for child items which is not correct in either case as it always depends on the parent item.

marble:
I'm not aware of the actual recommendations we're having right now in our "Writing Documentation" manual. But I have a strong opinion on how we should deal with it. The GENERAL procedure would be like follows. It is more a general attitude we should have:

- We should ACCEPT every documentation from a contributor that technically works.
- We should not edit reST sources for style reasons only.
- We should have a definition in a "T3Docs ReST Style Guide" of "desired target states." [^target-states]
- So the rules in the style guide is what we tell people when they are asking. And updates done by different people should converge. And we want to get rid of having the same discussions all over again.

[^target-state]: For example, 1 tab = 3 spaces is such a desired target state. The idea is, that style changes of different people should converge. Otherwise they'd be changing things back and forth. Whether people WILL use or HAVE used different styles is irrelevant for the process of defining a desired target state. 

A bit more informal:
* For example, a line should only have 1 to 79 characters plus NL character in it. That is important to agree on and to know because then you can set that value in plugin "Wrap plus" of Sublime that does "intelligent" reformatting of whole paragraphs. 

It's irrelevant whether people have used the 79 limit up to now.

* It is important to go for `*  Star Blank Blank at the beginning` of a list because then you can use the TAB-key to indent all the lines of a list item if it spans several lines. Additionally it makes visual appearance more consistent and stress the idea of a "senkrechte Fluchtlinie" (visual vertical helper line) between column 3 and 4. Example: I had to change this page for a content reason. So I took the chancea and made the whole file constistent. Watch the result: https://docs.typo3.org/_sources/Home/WhatsNew.rst.txt

* Pages and manuals can be very different and there may be good reasons to do something differently. For example: "In general" I use ONE empty line to separate things. BEFORE a headline I use TWO empty lines. Sometimes I find it's necessary to have THREE empty lines because otherwise it's to hard to find things in the text. I don't use more the three. But: If I detect, that somebody else is using consistently something else like 10 empty lines in a manual that S/HE maintaines I would respect that as well.



*

alexander-nitsche:
Ok. So how to proceed here: Which parts of the PR should be merged and which not?

alexander-nitsche:
hi @marble , hi @sypets , any opinion on how to proceed here? greetings, alex

marble (approved):
I'm absolutely PRO using the three chars indent for unordered lists as stated in the comments. Usually we should reformat just for the sake of reformatting. But since this PR now is there and looks good to me - please check it in. Leaving that for Sybille, as I see her as the "maintainer".

sypets:
Maybe we should discuss the general stuff elsewhere (perhaps on decisions.typo3.org) - I think this is a very general topic - what formatting / CGL should be used, should formatting be enforced etc. ... 

Martin:
> We should ACCEPT every documentation from a contributor that technically works.
> We should not edit reST sources for style reasons only. 

A reason might be to not make it too difficult for contributors and not create extra work for maintainers.

But I think this creates more extra work in the long run. Because at some point the inconsistencies should get fixed. **And also - people keep using the "wrong" source as basis - so it gets repeated.**  


> There will never be a tool that enforces a single set of rules on the whole text. That would be too much of a restriction for humans. 

In the core, there is - see coding guidelines. There are just not rules for everything, for some things you cannot apply "rules" or it is not possible to automatically detect if the text is correct. That is the case for language issues. But even for that, you have spell checkers. You can at least look at the warnings. 

That does not mean that a tool it has to or can cover everything and reject everything. For example, for the line length - that is rather a recommendation. You can put it in .editorconfig but you don't have to let tests fail if it is not applied (line length in commit messages are enforced though, in the core by the commit hook). But for indenting, you could enforce and convert it (if such a tool existed).

If we had **one way to do things** (at least for some things) - I think it would make it a lot easier - that concerns CGL, but also how to format things (there are several ways to create external links, code blocks), the spelling, how to structure things on a page (e.g. add menu on top?), how to structure a chapter (have entry page with only menu), when to use the 1, 2, 3 lists, should linktargets be added to menu, where should the sitemap be (top of menu, bottom), how to create images (with shadow, which icons to use), the general structure, link to changelogs or not ... etc. etc. 


This is how it works in the core:

* it is strongly encouraged to fix patches from new contributors with CGL issues for them, see https://docs.typo3.org/m/typo3/guide-contributionworkflow/master/en-us/HandlingAPatch/Tips.html#be-helpful
* no core patch with CGL issues is merged - it is impossible because typo3.com / bamboo gives it a -1
* typical IDEs can enforce (most of ) the coding styles (PSR-1, PSR-2, .editorconfig)
* there is an additional script which can convert the source for CGL (at least for PHP)
* there is even a commit hook which checks this
* CGL - among other things - are checked via automatic tests with bamboo

IMHO, it would be great to move in the direction of the core. But as long as the tools are not available - I would personally not invest _too_ much time and effort into this. But I think it is also a good idea to already decide things and have a recommended way even if it cannot be fully enforced currently. 

For me, having no rules or not enforced rules in the docs (not just CGL, but other stuff too) **made it not easier to contribute but very difficult to contribute**. For every bigger change (that is not a simple fixing of a typo) I can into unclarified or inconsistent things or things where it was not clear how it should get done,  where I was not sure how to do them - that concerns icons in images, coding guidelines, spelling of titles, how content is structured, common terms with common spellings etc. etc.

I tried to clarify a lot of this, create decisions, write it down here, etc. but it is really very tedious. And after a while it is not a lot of fun. I think it would be good to get some feedback on this - if others feel the same way. (My reasoning was, clarify it once - then you don't have to keep discussing it).

That makes it overall easier for me to contribute in the core, despite the fact that writing documentation would actually be a lot easier for me than writing (good enough for the core) code.

The core tries to set the bar high - with enforced CGL, phpstan, unit tests etc. - this helps to ensure quality. I think the docs should have a similar bar - of course, not just of these formal things but of the content in general. This might make it more difficult to contribute for some people, but others can help with that. And things that are difficult in the beginning, may cease to be difficult after a while. The contributing will be much more rewarding because you know you are participating in something of high quality. In the past this was not always the case.

This is an almost philosophical question - but my point is - trying to make it easier for people might not really be making it easier!

alexander-nitsche:
I agree 100% with you, Sybille: Having a clean way for every documentation task would be my quo-vadis. Applying your thoughts to this issue: I would always rework the whole repository if a new rule has been defined such that every reader can find arbitrary how-tos by just looking at the reST files of it. Even reworking whole repositories becomes easier and easier the more consistent the reST structure is.

marble:
I don't think we disagree. Maybe I tend to be a bit more pragmatic, but probably it's not even that. I'm probably more pessimistic that there'll be a short and understandable ruleset to fit them all. And I don't want to repell beginners. 

Let me find out what I think by writing here:

We should use good, defined and consistent style. If in doubt, I'd rather tolerate violations of the recommendations than rejecting input from somebody. If new docs are created I'd rather have them styled well. It would be desirable to adapt existing manuals to our rules in a whole. Otherwise document by document, it it's updated anyway. I wouldn't like that changes from ordinary people get rejected only because of style reasons.

And I assume that there will be much more "should have" rules for documentation than for the core. For example, personally I'd vote for: "There should be a blank line between list items."

But if it then happens that I have a document with many list with short but many list items, I would understand why somebody would like them to write as:

     *  Something short abc
     *  Something short bcd
     *  Something short cde
     *  Something short def
     *  Something short efg
     *  Something short fgh

    And another one:

     *  Something short abc
     *  Something short bcd
     *  Something short cde
     *  Something short def
     *  Something short efg
     *  Something short fgh

I have to confess that I'm aware of what recommendations you are giving in the Writing Documentation manual. That's not disrespect, just lack of time. Should we have a standalone TYPO3-REST-STYLEGUIDE.rst file, similar to a README.rst file, just listing the rules? Like:

    * Start writing in the first line of a document.
    * Use UTF-8. This means, ASCII is ok as well as it's a subset.
    * Don't use tabs, use blanks instead for indentation.

    and so on.

alexander-nitsche:
Hi Martin, i understand your point of not rejecting newcomers and facilitate participation. I guess this could be handled as in the core team and a senior editor just fixes the formatting once, gives hints to the newcomer and merges the changes instead of rejecting them. As for the rules i guess there are already three places to maintain them: the .editorconfig and the docs chapter https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/Index.html and the documentation reST code itself. Maybe we add there recommendations and requirements regarding specific reST elements (like e.g. lists). Greetings, Alex

sypets:
I will merge now. I think we should add a (short) explanation and link to the issue #176 where there is a longer explanation, but do that in a followup PR.

PR #184, opened 2020-12-10
Start with index page

marble:


marble:
This is ready to go. Just push!
grafik

Once enough index entries are added I would move the "Index" item up to to top of the menu.

linawolf (approved):

PR #187, opened 2020-12-22
Write ViewHelper in CamelCase

linawolf:
According to here: https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/GeneralConventions/ContentStyleGuide.html#exceptions-for-specific-typo3-spellings and most other documentation ViewHelper should be written in CamelCase. If it was not it would have to be written in two word: view helper.  What do you think?

### DIFF
--- a/Documentation/GeneralConventions/Glossary.rst
+++ b/Documentation/GeneralConventions/Glossary.rst
@@ -241,7 +241,7 @@ TypoScript
 Spelling:
    :ref:`TYPO3 exception <spelling-typo3>`
 
-Viewhelper
+ViewHelper
 ----------
 
 


alexanderschnitzler:
I agree. Everybody working or writing about TYPO3 uses `ViewHelper`. It's a fixed term. `Viewhelper` is an option but it's a distraction in texts. Pro `ViewHelper`.

alexanderschnitzler (approved):

marble:
https://typo3.slack.com/archives/C028JEPJL/p1608658455207000

Hey, do we say ViewHelper or Viewhelper?

marble:
I would go for the easiest and most natural form:
view
helper
view helper
+ proper names

Example:
""""View helpers - yes, TYPO3 has a lot of those view helpers. As there are: BaseViewHelper, DebugViewHelper and many more. They usually live in a folder :file:'ViewHelpers' (should be backticks).

In the example Oracle page: https://www.oracle.com/java/technologies/viewhelper.html it's just "view" and "helper" too. Except that they coin the special term "View Helper pattern" whicht they always write like that.

Following our newest "insights" at TYPO3 Explained we could write this as (backticks!) 'View Helper pattern':pn: meaning that is has the "proper name" (Eigenname) text role. See discussion over there.




"""
"""

marble:
It is ok with me to just write "ViewHelper". The reason I thought it was meant to be different is this new entry in "TYPO3 Explained": 
image

marble (approved):
Please read my comments in #187

PR #190, opened 2020-12-23
Update Glossary.rst, add text role 'pn'

marble:
Explain text role 'pn' and improve usability of page "Spellings, terms and glossary"

image


linawolf (approved):

linawolf:
I think backend is not a proper noun or it would have been written in UpperCase?

marble:
> I think backend is not a proper noun or it would have been written in UpperCase?

Agreed, sure. But it may still be in the list just to show how we spell words. It is not "back end", as somebody new an naiv might think. However, we wouldn't mark that up as "proper name".

PR #196, opened 2020-12-25
Add indexes to HowToDocument

linawolf:
and update headings to sentence case
Resolves: #192


sypets:
Several unfinished indexes have been added in several manuals. Can we do this one at a time and finish one index before the next one is started? That way we do not have too many unfinished pages around.

(There is another one in TSconfig and possibly several others).

We cleaned up a lot of unfinished "to be written" pages and updated and improved. I think adding unfinished pages is a step in the wrong direction.

Also, I think the index might be helpful but more important would be working on the structure in general and have a plan. We had a Slack call about that a while ago, we agreed to work on the structure.

We added an index to the "Contribution Guide" about a year ago. Is it being used? Is it helpful? Are all maintainers and contributors on board for  maintaining the index as well?

linawolf:
This index is finished, it is just waiting for review. Same goes with the ones in the other manuals. And it is an excellent tool in trying to find duplicate information and restructuring content. You can see a finished one in TYPO3 explained:
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/genindex.html

marble:
> Several unfinished indexes have been added in several manuals. 

I find your comment destructive, annoying and demotivating. Why do you take the effort of writing such a long comment instead of doing the review of the respective PR? And, please, don't tell other people - again - what they should be doing better according to your opinion. I don't seen anybody else complaining about the emerging index pages. If you don't like them - look somewhere else. But don't ruin other people's motivation.

marble (approved):

PR #199, opened 2020-12-26
Unify link targets text

sypets:


linawolf (approved):
added indexes and rewording to plural.

PR #204, opened 2021-01-04
Fix typo

marble:


linawolf (approved):
They write themselves YouTube, in both logo and text  https://en.wikipedia.org/wiki/YouTube

marble:
@linawolf Yes, that's why I made this mistake. Because: In this context you are addressing the reST DIRECTIVE - and all those directives you have written as they are, that is, 'youtube'. Maybe we should find a way to express that we are referring to the literal form of a word in those cases. We could write: `.. index:: reST directive; 'youtube'`. As there are
image
and so on

linawolf:
However the sorting be by letter on the first level only works, if there are not special signs

PR #209, opened 2021-01-10
Example 3: List with sublist and whitespace error

linawolf:


gilbertsoft (approved):

marble:
- What's bad with this page is that it is not very obvious that examples 2 and 3 are actually bad ones that are to avoid. Difficult to understand.

- If available for your editor, install a plugin like "Rainbow indent" for PhpStorm and Jetbrains IDEs). I find that helpful.

- My proposal: Demonstrate possible errors by showing the source screenshots. I have something in mind like the following. They just need some arrows and annotations:

### BAD

1. Bad: Line not matching indentation level (item 1)
2. Bad: One or more empty lines missing before sublist (before subitem 3.1)
3. Bad: Line not matching indentation level (subitem 3.1)


Annotations:

### GOOD

PR #235, opened 2021-09-30
[TASK] Disable all dummy urls with `:samp:`

alexander-nitsche:
See TYPO3-Documentation/T3DocTeam#164 for details.

### DIFF
--- a/Documentation/WritingReST/Hyperlinks.rst
+++ b/Documentation/WritingReST/Hyperlinks.rst
@@ -58,6 +58,13 @@ by adding a label before the section header::
 
 How to create links is described in more detail in the next sections.
 
+:ref:`Preventing links <preventing-links>`
+   being automatically generated from simple URLs.
+
+   .. code-block:: rest
+
+      :samp:`<url>`
+
 
 .. index:: reST; External links
 .. _external-links:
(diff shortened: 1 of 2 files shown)


Documentation/WritingReST/Hyperlinks.rst
```diff
@@ -58,6 +58,13 @@ by adding a label before the section header::
 
 How to create links is described in more detail in the next sections.
 
+:ref:`Preventing links <preventing-links>`
```
mbrodala, line 61: IMO this shouldn't be the first thing to mention in a section about linking. ;-)

Since we already have a more thorough description at the end of this section we can skip it here.

alexander-nitsche: To me the first section looked like a Table of Contents. If you use it like links to most important parts of the further documentation, then we should skip it, yes.

mbrodala: Ah indeed, just had another look at the rendered result. Still we should move this after the 3 existing entries here, again because it's less important than actual linking.

alexander-nitsche: yes, it should be already at the last position ;)

mbrodala: I already had my coffee, I swear. Thanks for pointing out the obvious to me. :+1:

mbrodala (approved):

PR #244, opened 2021-11-24
[TASK] Unify placeholder syntax

alexander-nitsche:
See TYPO3-Documentation/T3DocTeam#166 for details.


sypets:
Can we add ":samp:" to this page as well? https://docs.typo3.org/m/typo3/docs-how-to-document/master/en-us/WritingReST/InlineCode.html That is where I first looked and couldn't find it.

susannemoog (approved):
@sypets that should be done in a separate issue I think.

PR #245, opened 2021-11-25
[TASK] Unify placeholder syntax - Documentation

alexander-nitsche:
See TYPO3-Documentation/T3DocTeam#166 for details.

### DIFF
--- a/Documentation/WritingReST/CheatSheet.rst
+++ b/Documentation/WritingReST/CheatSheet.rst
@@ -183,7 +183,7 @@ and emphasize parts of the URL with curly braces:
 .. code-block:: rest
 
    The *route* is the "speaking URL" as a whole without the domain part,
-   for example :samp:`https://example.org{/unlinked-urls}`.
+   for example :samp:`https://example.org/<page-path>/{<page-name>}`.
 
 .. index:: reST; Hypothetical domains
 .. _cheat-sheet-hypothetical-domains:
@@ -335,6 +335,21 @@ Source::
    Assign the variable a::
 
       $a = 'hello';
+
+
+Placeholders
+------------
+
+Use the *Backus-Naur form* `<placeholder-name>` for placeholders in code and
+URLs.
+
+Examples:
+
+#. :php:`class <Entity>Controller extends ActionController`
+#. `typo3/sysext/core/bin/typo3 impexp:import <file>`
+#. :samp:`https://example.org/typo3/main?token=<token-id>`
+
+For XML and HTML, use the comment tag :html:`<!-- placeholder-name -->`.
 
 
 .. index:: reST roles
--- a/Documentation/WritingReST/Codeblocks.rst
+++ b/Documentation/WritingReST/Codeblocks.rst
@@ -24,6 +24,9 @@ Quick Reference
 *  It is recommended to use the short form (`::`) instead of code-block explicitly.
 *  Always use :ref:`syntactically correct code <codeblocks-syntactically-correct>`
    in a code block.
+*  Use :ref:`placeholders <codeblocks-placeholders>` in angle brackets
+   (`<placeholder-name>`) to refer to a place in the code where you don't care
+   about the exact value.
 
 The following examples all do the same thing:
 
@@ -836,3 +839,55 @@ Literalinclude
 
 There also is a `literalinclude directive
 <http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-literalinclude>`__.
+
+
+.. _codeblocks-placeholders:
+
+Placeholders
+============
+
+Placeholders in this context are named tags in code and
+:ref:`example URLs <preventing-links>` where the exact value does not matter,
+but is referenced in the surrounding documentation.
+Use the *Backus-Naur form* `<placeholder-name>` for placeholders in code and
+URLs, i.e. use angle brackets to encapsulate the placeholder name.
+
+For example in PHP
+
+.. code-block:: rst
+
+   Set up a controller class to handle user interaction with the entity data
+   model:
+
+   .. code-block:: php
+
+      class <Entity>Controller extends ActionController
+      {
+         ..
+      }
+
+   where `<Entity>` corresponds to the entity data model class name.
+
+or on the command line
+
+.. code-block:: rst
+
+   Importing a TYPO3 dump file is as simple as running:
+
+   .. code-block:: bash
+
+      typo3/sysext/core/bin/typo3 impexp:import <file>
+
+   where `<file>` can be the absolute path on the server or the relative path
+   in the TYPO3 project folder.
+
+or in an example URL
+
+.. code-block:: rst
+
+   The TYPO3 backend normally appends the session token ID to the URL as
+   follows: :samp:`https://example.org/typo3/main?token=<token-id>`.
+
+In the XML and HTML markup languages, which make extensive use of angle
+brackets, the comment tag :html:`<!-- placeholder-name -->` is used to insert
+placeholders.
--- a/Documentation/WritingReST/Hyperlinks.rst
+++ b/Documentation/WritingReST/Hyperlinks.rst
@@ -328,9 +328,11 @@ Preventing links
 
 Sphinx automatically converts simple URLs into links. This can be unintentional
 in certain contexts, for example when using a
-:ref:`hypothetical domain <hypothetical-domains>` like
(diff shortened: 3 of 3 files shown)


Documentation/WritingReST/CheatSheet.rst
```diff
@@ -183,7 +183,7 @@ and emphasize parts of the URL with curly braces:
 .. code-block:: rest
 
    The *route* is the "speaking URL" as a whole without the domain part,
-   for example :samp:`https://example.org{/unlinked-urls}`.
+   for example :samp:`https://example.org/<page-path>/{<page-name>}`.
```
mbrodala, line 186:
```suggestion
   for example :samp:`https://example.org/<page-path>`.
```

Adding another segment to the whole path looks confusing IMO.

Documentation/WritingReST/CheatSheet.rst
```diff
@@ -337,6 +337,21 @@ Source::
       $a = 'hello';
 
 
+Placeholders
+------------
+
+Use the *Backus-Naur form* `<placeholder-name>` for placeholders in code and
```
mbrodala, line 343:
```suggestion
Use the `Backus-Naur form<https://en.wikipedia.org/wiki/Backus-Naur_form>`__ like `<placeholder-name>` for placeholders in code and
```

Documentation/WritingReST/CheatSheet.rst
```diff
@@ -337,6 +337,21 @@ Source::
...
+URLs.
+
+Examples:
+
+#. :php:`class <Entity>Controller extends ActionController`
```
mbrodala, line 348: IMO we should remove this since we'd usually use a full and obvious name like `ExampleController` instead. Also In PHP code this looks like [generics](https://wiki.php.net/rfc/generics) and thus even more confusing.

Documentation/WritingReST/CheatSheet.rst
```diff
@@ -337,6 +337,21 @@ Source::
...
+#. `typo3/sysext/core/bin/typo3 impexp:import <file>`
+#. :samp:`https://example.org/typo3/main?token=<token-id>`
+
+For XML and HTML, use the comment tag :html:`<!-- placeholder-name -->`.
```
mbrodala, line 352:
```suggestion
For XML and HTML, use the comment tag :html:`<!-- placeholder-name -->` since `<placeholder-name>` looks like a regular element.
```

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: I try to reduce external links. One of my guidelines is to keep the reader on the page or at least in the documentation or TYOP3 universe, and do not try him to leave. If he really wants to look up the name, (s)he should do a copy&paste of the name.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: Ok, if this looks to much like generics, but instead of removing could you replace it with another PHP code example?

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: In the cheat sheet i would list only the shortest version of the rules and move any explanation to the detailed chapter.

Documentation/WritingReST/CheatSheet.rst
mbrodala: And I try to add as much explanation as possible, using external links if possible. This way the reader has a curated resource to look at. He is free to skip the link if not necessary.

If you prefer to keep people in the TYPO3 docs then we should add a short explanation what BNF is. This way one can continue reading without being puzzled or having to read an external explanation which would interrupt the reading flow.

Documentation/WritingReST/CheatSheet.rst
mbrodala: Sure, e.g. said `ExampleController`.

Documentation/WritingReST/CheatSheet.rst
mbrodala: Fine with me, wherever it makes the most sense.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: > He is free to skip the link if not necessary.

Yes, but he gets distracted with every linked word. On the other hand i would not know which part of the BNF explanation would be required to further understand the way we want to highlight the placeholder.

I don't mind too much, i see that there two ways to approach linking of external sources, so i would suggest to discuss it in the TYPO3 Documentation Team and come up with a guideline.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: Yes, but there is no placeholder in that example then.

Documentation/WritingReST/CheatSheet.rst
mbrodala: Suggestion to simplify this: drop the term _Backus-Naur form_ altogether. What is not mentioned does not need any explanation. It's just a syntax we chose to indicate placeholders/variables, that it's based on BNF is actually not really important.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: Yes, it is already in the detailed chapter, see file CodeBlocks.rst:

> In the XML and HTML markup languages, which make extensive use of angle
brackets, the comment tag :html:`<!-- placeholder-name -->` is used to insert
placeholders.

Documentation/WritingReST/CheatSheet.rst
mbrodala: Correct, because it does not make sense to have one here. Code does rarely use placeholders because it already has something similar: variables. Or static identifiers like class names. These by themselves already serve as placeholders.

Documentation/WritingReST/CheatSheet.rst
mbrodala: Then we should add my suggestion there. I always try to add a justification for rules.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: So you say, we should not add a PHP example as there are no ways to use placeholders in PHP in a useful way? I think, the whole Convention-over-Configuration setting in Extbase is a very good example of using placeholders. To explain the naming schema of a model, its controller and its repository, you can either use an real world example or just describe it in this templating way. And to not confuse the reader you can pick up the placeholder in your text flow and tell what it is about.

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: Ok, would you suggest your change there?

Documentation/WritingReST/CheatSheet.rst
alexander-nitsche: I would like to keep it in the text, as it is an explanation of why we have chosen it (as it is following some standard) and would like to discuss the external topic link in the team.

Documentation/WritingReST/CheatSheet.rst
mbrodala: As discussed we can keep this as example. We'll see if it really leads to confusion.

Documentation/WritingReST/Codeblocks.rst
```diff
@@ -836,3 +839,55 @@ Literalinclude
...
+or in an example URL
+
+.. code-block:: rst
+
+   The TYPO3 backend normally appends the session token ID to the URL as
+   follows: :samp:`https://example.org/typo3/main?token=<token-id>`.
+
+In the XML and HTML markup languages, which make extensive use of angle
+brackets, the comment tag :html:`<!-- placeholder-name -->` is used to insert
+placeholders.
```
mbrodala, line 893:
```suggestion
placeholders. A `<placeholder-name>` looks like a regular element and would lead to confusion.
```

Documentation/WritingReST/CheatSheet.rst
mbrodala: Done.

mbrodala (approved):

mbrodala:
Thanks for the patience.

alexander-nitsche:
Thanks for your company, i enjoyed it.

PR #261, opened 2021-12-22
[TASK] Add signposting to landing page

### LINKED ISSUE #152: Feedback from Twitter: create one pager quickstart into TYPO3 rst docs for Markdown users and advertise this prominently in different places.
> my suggestion would be to create a one pager quickstart into TYPO3 rst docs for Markdown users and advertise this prominently in different places.

https://twitter.com/helhum/status/1246061661699690496

tomwarwick:
Resolves #152

Following community feedback, this commit adds
links to the different sections of this guide on the
homepage along with a brief explanation for each
chapter.

The intent of this commit is to highlight the different
types of documentation that can be created. 
This includes content for extensions, sys extensions,
third party exts and the big one - docs.typo3.org


sypets (commented):
* The headlines look a little big to me. We usually do not have headlines at all on start pages
* Can we use the card layout here too?
* Why use bullets for only one bullet point?
* Can we make decision how to structure start page in general?

Apart from that, I think it is a good idea to have the sections for the various things to do - so from a content perspective, I think this is a good change. I would just rethink the formatting.

susannemoog (approved):

Issue #255, opened 2021-12-18
Document confval directive 


sypets:
`.. confval::`

And clarify usage - for what should it be used, e.g. references of functions? References in general?

For example, in the rte_ckeditor documentation, the `.. option::` is used. The result looks similar:

* source: https://docs.typo3.org/c/typo3/cms-rte-ckeditor/main/en-us/_sources/Configuration/Reference.rst.txt
* docs: https://docs.typo3.org/c/typo3/cms-rte-ckeditor/main/en-us/Configuration/Reference.html#editor-config

sypets:
Has been solved with PR https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/262

Issue #254, opened 2021-12-18
Document use of card layout for start pages


sypets:
see also https://github.com/TYPO3-Documentation/DocsTypo3Org-Homepage/pull/182

sypets:
Resolved with https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/262

PR #262, opened 2021-12-31
[FEATURE] Document diverse ReST content elements

linawolf:
* Cards
* Tabs
* Sidebar
* Confval
* Phpdomain

solves #255
solves #254


Documentation/WritingReST/Index.rst
```diff
@@ -45,13 +45,12 @@ http://openalea.gforge.inria.fr/doc/openalea/doc/_build/html/source/sphinx/rest_
...
 
 
-
-
 .. toctree::
    :hidden:
 
    Introduction
    BasicRestSyntax
+   ContentElements/Index
```
alexander-nitsche, line 53: Adding the new content elements as subpages of the new folder "ContentElements" is breaking the current organizing, where each content element is added as subpage to "Reference" and as content element to page "CheatSheet". Will continue with reviewing the single pages as soon as the reorganizing is done.

alexander-nitsche (changes requested):

alexander-nitsche (approved):
Ready for merging from my side.

tomwarwick (approved):

sypets:
@linawolf tip: If using for **Resolves** with a colon to refer to the issue, for example

`Resolves: #255`

as in TYPO3 core - this will close the issue automatically. Here, `solves #255` was used.

(I think you can also use Fixes: but I think it would be easier for everyone if we use the same conventions as in TYPO3 core contribution).

If you do **not** want the issue to close automatically (yet), you can use **Related** instead of Resolves. This is the same in the TYPO3 core. Using this in both core and documentation should hopefully make it easier to manage the issues and get them closed automatically.

This is also described int https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/CommitMessages.html

PR #266, opened 2022-03-13
[TASK] Revision of the page "Directory and file names"

alexander-nitsche:
Some texts were outdated and the structuring of the sections
needed to be improved to be better linkable from the outside.

This new page "File structure" should provide all information
for a newbie regarding file structure and file content of a
TYPO3 documentation, including templates and examples.

Related issue: https://github.com/TYPO3-Documentation/T3DocTeam/issues/167
Related issue: https://github.com/TYPO3-Documentation/T3DocTeam/issues/168


Documentation/GeneralConventions/FileStructure.rst
```diff
@@ -0,0 +1,810 @@
...
+
+.. _examples-typo3-third-party-extension-documentation:
+
+TYPO3 third-party extension documentation
+-----------------------------------------
+
+.. table::
+   :widths: 25, 75
+
+   ========================  ===================================================
```
linawolf, line 776: I think we should get rid of these tables and not introduce new ones. They break easily and do not display well on smaller screens. We could convey the same information with definition lists or headlines or some such

Documentation/GeneralConventions/FileStructure.rst
```diff
@@ -0,0 +1,810 @@
...
+
+   .
+   ├── README.md
+   └── Documentation
+       └── Settings.cfg
+
+
+.. _single-file-documentation-readme-md:
+
+README.md
```
linawolf, line 646: We are using rst all over. Why should we suggest mark down here even though rst is possible?

Documentation/GeneralConventions/FileStructure.rst
```diff
@@ -0,0 +1,810 @@
...
+project with as many keywords as possible in as few sentences as possible. It
+helps the decision maker to quickly decide whether the project is worth
+considering and whether or not to read the full documentation. It should be
+aligned with the abstract of Index.rst and - if available - the description
+fields of ext_emconf.php and composer.json.
+
+
+.. _index-rst:
+
+Documentation/Index.rst
```
linawolf, line 167: I think this is not really the best content for a index.rst a much more modern aproach would be a tile structure like we have in TYPO3CMS-Guide-HowToDocument or TYPO3CMS-Tutorial-GettingStarted.

This strange outdated table like structures makes ppl afraid to use the start page for something creative. Beside the fact that we need a toctree (which can be hidden) we really dont need this.

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: They probably break for more columns or for longer words. For two columns and short words they work fine:

Screen Shot 2022-03-14 at 15 55 16

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: Yes, as the documentation is also for TYPO3 extension documentations, where many like to keep the README.md, i think it is fine to provide examples for both cases here.

linawolf: I would prefer to suggest to new comers to use rst and then just mention (and give an example) that you can also keep the README.md instead of rst.

Documentation/GeneralConventions/FileStructure.rst
linawolf: The information we provide here isn't really necessarily  of tabular nature so I would not use tables for it.

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: Ok, will invert the order.

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: The meta data table contains duplicate information, but it has been greatly reduced, and there is also unique information such as the extension key/package name and authors. It also contains the unique abstract and a unique view of the expanded menu. Finally, all this information is presented in the main content area, so the reader has it at hand immediately. Especially on mobile devices, it makes a difference to put meta data somewhere in the frame of the page, e.g. at the very bottom of the footer, or at the very top of the content area.

As the academic style is used in most current documentations and i can also see use cases for it for coming documentations, especially in extension documentations, i want to establish a standard for it (this was my motivation, in fact). This does not keep you from (1) writing down a standard for Index.rst in this portal style and (2) not from applying it in all your documentations.

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: So this is about taste. Which content element would you prefer?

Documentation/GeneralConventions/FileStructure.rst
alexander-nitsche: Done.

alexander-nitsche:
The new "File structure" page is ready to replace the "Directories and file names" page and to be the central point for further explanations of each file of a basic TYPO3 documentation.

Documentation/GeneralConventions/FileStructure.rst
linawolf: I would probably have solved it with a definition List or another List of some sorts. If I would introduce a table I would prefer one of the easier extendable formats like https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Tables.html#csv-table however, the discussion on weather we get rid of the first two table formats is out of scope here. So I am ok with your tables for now, we can always change them later on.

linawolf (approved):

PR #267, opened 2022-03-17
[TASK] Use "Key" and "Package" fields in Index.rst template

alexander-nitsche:
Fields with spaces don't look very nice when wrapped due to
extended field values, so rename the "Extension key" field
to "Key".

In addition to the TYPO3 extension key, the Index.rst should
also list the name of the project's Composer package - if any -
and use the new "Package" field for that.

Follow-up of #266


mbrodala:
Can you link an example / embed a screenshot of what you mean?

People are used to the terms _Extension key_ and _Package name_ so dropping them is quite some major change.

alexander-nitsche:
Both terms are barely used currently, whereas the old keyword was "Classification" which is far worse :)

### Screenshot
Screen Shot 2022-03-17 at 16 40 25

alexander-nitsche:
With "Extension key" and "Package name":
Screen Shot 2022-03-17 at 16 44 01

alexander-nitsche:
It seems that the field names are not wrapped anymore due to the shorter license text now. So i would tend to rename the fields again to "Extension key" and "Package name".

alexander-nitsche:
Reverted field names to "Extension key" and "Package name". The remainder of the change remained.

Documentation/GeneralConventions/FileStructure.rst
```diff
@@ -246,17 +249,18 @@ Extension key
...
+Package name
+^^^^^^^^^^^^
 
-and insert the Composer package name into the *package-name* placeholder.
+The *package-name* placeholder contains the Composer package name of the
+project.
 
-Remove this table row if the documentation does not refer to either a TYPO3
-extension or a PHP application.
+Remove this field if the project has no package name.
```
mbrodala, line 263: That's basically impossible, since the package cannot be installed then.

alexander-nitsche: This guide is also for official TYPO3 manuals, like e.g. the "TCA Reference".

mbrodala: Then:

> Remove this field if the project is no installable package

alexander-nitsche: Done.

mbrodala (approved):

PR #268, opened 2022-03-20
Lets not use :ts: for typoscript

linawolf:
In code blocks `.. code-block:: ts` is used for typescript as a sphinx standard. code-blocks with code-type ts may therefore not contain typoscript or code highlighting will fail. the practice of using `:ts:` for inline code-blocks creates an unnecessary confusion as ppl will then also use it for code-blocks. This has caused missing syntax highlighting and rendering warnings throughout the different documentations.

Unfortunary I didn't catch this in the first review of this document

### DIFF
--- a/Documentation/GeneralConventions/FileStructure.rst
+++ b/Documentation/GeneralConventions/FileStructure.rst
@@ -403,8 +403,6 @@ markup languages used in a TYPO3 project:
    .. role:: rst(code)
    .. role:: sep(strong)
    .. role:: sql(code)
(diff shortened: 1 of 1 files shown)


alexander-nitsche:
sounds good, but requires scanning and replacing of current use of `:ts:` in all existing repos and branches, not?

alexander-nitsche (approved):
Nevertheless it is the right thing to do.

linawolf:
we fix all rendering warnings anyway when applying these rules? and its a simple search and replace

PR #270, opened 2022-03-27
Add missing reference skinning

linawolf:
to standard Settings.cfg


alexander-nitsche:
Is this manual "TYPO3 Skinning" still maintained for the current LTS?

linawolf:
No but it seems like we should leave a warning like in the other outdated manuals.

alexander-nitsche:
Ok. Let's continue with archiving that manual instead.

PR #272, opened 2022-04-05
Update Inline code

marble:
* Add `bash` and `shell` textroles for `code`
* Improve color contrast for all inline styles derived from `code`

### DIFF
--- a/Documentation/Includes.rst.txt
+++ b/Documentation/Includes.rst.txt
@@ -7,11 +7,13 @@
 
 .. role:: aspect(emphasis)
 .. role:: bash(code)
+.. role:: css(code)
 .. role:: html(code)
 .. role:: js(code)
 .. role:: php(code)
 .. role:: rst(code)
 .. role:: sep(strong)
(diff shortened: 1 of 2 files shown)


marble:
The changes will really be visible only with the next version of the sphinx_typo3_theme, which will come with the next DRC (Docker Rendering Container)

alexander-nitsche:
hi martin, i would remove the `:shell:` text role again as the use case is a duplicate of that of `:bash:`. greetings, alex

fyi: in the current standardization run i replace all shell occurrences with bash.

linawolf:
I would support this PR as bash is a special case of a shell, the  Bourne-again shell. therefore `:shell:` is more generic then `:bash:`

alexander-nitsche:
putting my considerations in other words: with `bash` text role only the documentation writer immediately knows what text role to choose for his or her shell and bash code. with `bash` and `shell` text roles the writer looses this certainty and we will gain nothing.

linawolf (approved):

PR #306, opened 2022-09-17
[BUGFIX] Avoid quote style in numbered lists

brotkrueml:


Documentation/WritingReST/CheatSheet.rst
```diff
@@ -90,7 +90,7 @@ method 2: "External Hyperlink Targets":
 
     Check out more information on t3o_
 
-    _t3o: https://typo3.org
+    .. _t3o: https://typo3.org
```
linawolf, line 93: This was actually correct before not after the change: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/CheatSheet.html#external-links I would, however get rid of this syntax and just use a normal external link: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Links.html#external-links

brotkrueml: I removed the external hyperlink target in this PR. We should make a separate PR for this now.

I don't understand it: the current example is marked as warning by PhpStorm, Also I'm using the external hyperlink target extensively in my extension docs (with the two dots in front) which works correctly.

linawolf (approved):

PR #309, opened 2022-11-17
Add example for all admonitions

froemken:


brotkrueml (commented):
We want to avoid "Important" as it is rendered green (like tip and not red what you would expect), we mostly use "Attention" instead.

froemken:
Shall I change "important" to "attention" in my PR?

brotkrueml:
We have already a page with all admonitions: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Admonitions.html, So maybe a link to that page is enough.

Documentation/WritingReST/CheatSheet.rst
```diff
@@ -598,6 +598,8 @@ Source:
     text with a green box (as styled by our sphinx template).
     ---------------------------------------------------------------
 
+Visit :ref:`h2document:rest-admonitions` to see all allowed admonitions.
```
brotkrueml, line 601:
```suggestion
Visit :ref:`h2document:rest-admonitions` to see the available admonitions.
```

brotkrueml (approved):

PR #312, opened 2023-01-10
Document directives versionadded, deprecated in rst-reference

linawolf:


brotkrueml (commented):
We have it already here: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/GeneralConventions/CodingGuidelines.html?#how-to-add-version-hints

linawolf:
Yes and I am linking there in the seealso section. But it is missing in the rst-reference, we do not show how you can combine it with an admonition like note or warning and we do not show rendered results anywhere

brotkrueml (approved):

PR #319, opened 2023-06-13
Change Docker-Image in source command to the pulled one

dhoffmann1979:
The image used in the second step is now the pulled one from first step.


brotkrueml (approved):

brotkrueml:
@dhoffmann1979 Good catch! Thanks for the fix.

PR #322, opened 2023-07-16
Streamline rendering chapter

sypets:
- Remove page for rendering with docker-composer. This was never really officially supported.
- move content on subpage "Rendering with Docker" to main page in rendering chapter
- shorten and remove redundant text


linawolf (approved):
Thanks for these helpful changes

PR #324, opened 2023-08-04
Fixed How to render documentation numbered list by correcting indent

timothebot:
[How to render documentation](https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/RenderingDocs/Index.html) has a broken list. 
After the second step, the code block is shown as quote block and the next step starts at number one again, but should be number three.

This was solved by fixing the indent of the code block, which was previously three spaces instead of four.

### DIFF
--- a/Documentation/RenderingDocs/Index.rst
+++ b/Documentation/RenderingDocs/Index.rst
@@ -35,7 +35,7 @@ or :ref:`ask for help <get-help-on-writing-docs>`.
         The docker image is not listed on Docker Hub (hub.docker.com) anymore, therefore some preparations
         need to be done once:
 
-        .. code-block:: bash
+        ..  code-block:: bash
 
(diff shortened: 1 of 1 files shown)


Documentation/RenderingDocs/Index.rst
```diff
@@ -35,17 +35,17 @@ or :ref:`ask for help <get-help-on-writing-docs>`.
         The docker image is not listed on Docker Hub (hub.docker.com) anymore, therefore some preparations
         need to be done once:
 
-       .. code-block:: bash
+        .. code-block:: bash
```
brotkrueml, line 38:
```suggestion
        ..  code-block:: bash
```

brotkrueml (approved):

brotkrueml:
@timothebot Thanks for your contribution :+1:

PR #337, opened 2023-12-02
[TASK] Revise "How you can help" chapter

brotkrueml:


Documentation/WritingDocsOfficial/HowYouCanHelp.rst
```diff
@@ -83,189 +70,165 @@ cheat sheets.
...
+`#typo3-documentation`_ channel on `Slack`_.
 
-More Ideas?
-===========
 
-Do you have more ideas to add to this list? Either add them directly to this page using one of
-:ref:`the two workflows <contribute>`
-or `contact us
-<https://typo3.org/community/teams/documentation/#c9886>`__ and make your suggestions.
+..  _@documentation@typo3.social: https://typo3.social/@documentation
```
linawolf, line 226: I don't really like exchanging inline links for named ones. The inline links are much closer to html so I think easier to understand for devs. And also when you are reading the source code you can just open the link and don't have to look for it in another place. I think we also suggested somewhere to always use inline links. And we made effords to move away from the named links

linawolf (changes requested):
Please let us talk about it before switching already in place inline links to the old named link format

Documentation/WritingDocsOfficial/HowYouCanHelp.rst
brotkrueml: In this chapter I am using quite some links more than once. I wanted to avoid to add those multiple times. So I opted to move all the links to the end.

Additionally, I like the named links as the rst is better readable IMHO.

linawolf: It introduces more Syntax that we seldom use and have been trying to get rid of in the past years.

linawolf: It is implemented in the guides and what ppl are doing in extensions is their problem. But I think we should document a way to go and stick to it to make it easier for non-rst users to contribute. And I think we documented it here and should stick to that: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingReST/Reference/Content/Links.html#external-links If we want to use different syntax because it is very important then we should also document it there

brotkrueml: So I should add the same link over and over again on the same page, just to meet some other documentation?

fe-hicking: $0.02: I can understand both your POV. Especially because I don't really get RST links quite yet, I find those very hard to quickly understand and always shy away from it. Always looks like Perl code to me.

However.

Avoiding redundancy should IMO be the larger issue here, and to me it weighs over "requiring to learn other syntax". I'll have have to face that syntax at some point, and others writing docs probably too. RST already isn't that newbie-friendly...

linawolf: ok, lets keep it then, just don't make it a mass change all over then please

linawolf (approved):

PR #342, opened 2023-12-12
Correct rendering command

dhoffmann1979:
The rendering command prepared from the cli script of the image is named after the version of the image.


brotkrueml:
@marble Can you please have a look?

linawolf:
@dhoffmann1979 we would suggest, you migrate to the new, PHP based rendering container: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/Migration/Index.html 

The Sphinx-based rendering will be discontinued.

PR #347, opened 2024-01-13
[TASK] Adjust rendering page

linawolf:
Releases: draft


Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
+..  contents::
 
-This page explains how to render a manual locally on your machine
-with Docker in order to test the rendering.
+Rendering the :file:`Documentation` folder locally with Docker
+==============================================================
 
-Run these commands in a terminal in the parent directory of the directory
-:file:`Documentation`. You should use a bash compatible shell, if possible.
+You can render the documentation of an official TYPO3 manual or a third party
```
brotkrueml, line 22:
```suggestion
You can render the documentation of an official TYPO3 manual or a third-party
```

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
 
-.. rst-class:: bignums-xxl
+#.  Check if there is documentation to be rendered:
 
-    #.  Install Docker: https://docs.docker.com/install/
+    A folder called :file:`Documentation` containing at least the files
+    :file:`Index.rst` and :file:`guides.xml`.
 
-    #.  Preparations
+#.  Choose your prefered method of rendering (See bellow):
```
brotkrueml, line 34:
```suggestion
#.  Choose your prefered method of rendering (See below):
```

Additionally, where is below? Maybe link this to the section?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
+    ..  group-tab:: Linux
 
-            # The "real" tag is independent of the container repository,
-            # so let's just create that extra "real" and "generic" tag
-            docker tag ghcr.io/t3docs/render-documentation t3docs/render-documentation
+        ..  code-block:: bash
 
-            # verify it worked
-            docker run --rm t3docs/render-documentation --version
+            mkdir -p Documentation-GENERATED-temp
```
brotkrueml, line 44: Why one has to create the folder manually? I would expect that it is created automatically?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
-        the quotes around everything that goes into eval. Either define the function each
-        time you open a new terminal window or add the line to the startup file of your
-        shell like `~/.bashrc` or `.zshrc`.
+    ..  group-tab:: Windows
 
-    #.  Run dockrun_t3rd
+        ..  code-block:: powershell
 
-        .. code-block:: bash
+            start "Documentation-GENERATED-temp/Index.html"
```
brotkrueml, line 60: What does "start"?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
-:file:`Documentation`. You should use a bash compatible shell, if possible.
+You can render the documentation of an official TYPO3 manual or a third party
+manual with the following steps:
 
-If you run into a problem while rendering, check :ref:`rendering-docs-troubleshooting`,
-`report an issue <https://github.com/t3docs/docker-render-documentation/issues/new>`__
-or :ref:`ask for help <get-help-on-writing-docs>`.
+#.  Clone the repository containing the documentation.
 
+#.  Navigate to the directory, which contains the :file:`composer.json`
```
brotkrueml, line 27: Why composer.json and not just project root or similar?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
-    #.  Run dockrun_t3rd
+        ..  code-block:: powershell
 
-        .. code-block:: bash
+            start "Documentation-GENERATED-temp/Index.html"
+            docker run --rm --pull always -v ${PWD}:/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation
+            open "Documentation-GENERATED-temp/Index.html"
 
-            dockrun_t3rd makehtml
+Rendering extension documentation to docs.typo3.org
```
brotkrueml, line 64: I think "rendering" is the wrong word. Technically, on docs.typo3.org there is nothing rendered, it is the GHA. Maybe "publishing" is the better word?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
+===================================================
 
-        This will automatically find the documentation in the
-        :file:`Documentation` subfolder. It will create a directory
-        :file:`Documentation-GENERATED-temp` and write the results there.
+For your documentation to be rendered to our server https://docs.typo3.org, your
```
brotkrueml, line 67: rendered -> published ?

I would avoid "to our server". Who is meant with "our"? Is it obvious to the reader?
-> just "published on docs.typo3.org".

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
+TYPO3 extension has to have a valid :file:`composer.json` and either a folder
+called :file:`Documentation` with a :file:`Documentation/Index.rst` and
+a :file:`Documentation/guides.xml` or a :file:`README.rst` / :file:`README.md` on
+the level of your :file:`composer.json`.
```
brotkrueml, line 71: Again, why "composer.json"? It is the project/extension root?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
 
-    #.  Open generated documentation
+The extension has to be publicly available on GitHub or GitLab. You have to
+introduce a :ref:`Webhook <webhook>` and the Documentation Team has to
```
brotkrueml, line 74: "introduce" is the wrong word IMHO. Maybe "establish"?

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
+:ref:`approve <approval-intercept>` your first rendering request on a tool
+called "Intercept".
```
brotkrueml, line 76: Is Intercept relevant for the user? Can they do something with that knowledge? Maybe link to the docs rendering list, then it has a value.

brotkrueml (changes requested):

Documentation/RenderingDocs/Index.rst
linawolf: Right now it is not created automatically, once we have fixed the container so that it works without we can remove this line. Right now there is an error if the directory is not present

Documentation/RenderingDocs/Index.rst
linawolf: Just to give a reference what we are talking about. If for example you are in the TYPO3 monorepository you would have to be in the sysextensions folder - just where the composer.josn ist

Documentation/RenderingDocs/Index.rst
linawolf: It seems like on the mac it opens a browser and displays the file. The line was there before, I only adjusted the path. But I cannot verify it as I don't have a mac. Maybe @garvinhicking can help here

Documentation/RenderingDocs/Index.rst
linawolf: They can have a look at the public part of intercept. It is open for everyone

Documentation/RenderingDocs/Index.rst
linawolf: Ah sorry, lost context. It is a powershell command which opens the document in a browser

Documentation/RenderingDocs/Index.rst
linawolf: Ok, I removed it, we probably dont have to confuse ppl here

Documentation/RenderingDocs/Index.rst
linawolf: adjusted the text

Documentation/RenderingDocs/Index.rst
brotkrueml: But then it should be moved below the rendering command. But there is still the "open" command.

Documentation/RenderingDocs/Index.rst
```diff
@@ -10,97 +10,116 @@
...
-:file:`Documentation`. You should use a bash compatible shell, if possible.
+You can render the documentation of an official TYPO3 manual or a third-party
+manual with the following steps:
 
-If you run into a problem while rendering, check :ref:`rendering-docs-troubleshooting`,
-`report an issue <https://github.com/t3docs/docker-render-documentation/issues/new>`__
-or :ref:`ask for help <get-help-on-writing-docs>`.
+#.  Clone the repository containing the documentation.
 
+#.  Navigate to the extension root, the directory which contains the
```
brotkrueml, line 27:
```suggestion
#.  Navigate to the extension's root folder, the directory which contains the
```

PR #349, opened 2024-01-13
[TASK] Introduce automatic testing 

linawolf:
and remove all warnings


linawolf:
Merging this as it will cause to many conflicts

PR #352, opened 2024-01-14
[TASK] Drop the appendix

linawolf:
There is no more usefull information in here.


brotkrueml (commented):
Shouldn't this go to main as this is unrelated to the PHP-based rendering?

linawolf:
I thought to generally overhaul the guide and then do a switch. I don't think we should maintain two different versions..

brotkrueml:
> I thought to generally overhaul the guide and then do a switch. I don't think we should maintain two different versions..

Provide a PR against main branch, it gets merged, rebase documentation-draft.

linawolf:
See https://github.com/TYPO3-Documentation/TYPO3CMS-Guide-HowToDocument/pull/358

PR #363, opened 2024-01-15
[TASK] Remove the user Round trip

linawolf:
In my eyes it is not really helpful, to much information, graphs that dont really say much...


brotkrueml (approved):
Yup, does not belong to here.

PR #366, opened 2024-01-18
[TASK] Introduce backport workflow

linawolf:
So that while we are working on two versions
we can do automatic backports via label

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,16 @@
+name: Backport
+on:
+  pull_request_target:
+    types:
+      - closed
+      - labeled
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -0,0 +1,16 @@
...
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    runs-on: ubuntu-latest
+    name: Backport
+    steps:
+      - name: Backport
+        uses: m-kuhn/backport@v1.2.6
```
brotkrueml, line 14:
```suggestion
        uses: m-kuhn/backport@v1.2.7
```

PR #370, opened 2024-01-21
[TASK] Introduce Makefile and command 'make docs'

linawolf:
This is more convenient then having to copy-paste the command all the time

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -8,7 +8,7 @@ Information About Contributing to This Manual
 Local rendering
 ---------------
 
-You can render this file locally if you have `Docker <https://www.docker.com/>`
+You can render this manual locally if you have `Docker <https://www.docker.com/>`
 and `make <https://www.gnu.org/software/make/>` installed on your local
 machine::
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -5,6 +5,15 @@ Contributing
 Information About Contributing to This Manual
 =============================================
 
+Local rendering
+---------------
+
+You can render this file locally if you have `Docker <https://www.docker.com/>`
+and `make <https://www.gnu.org/software/make/>` installed on your local
+machine::
+
+    make docs
+
 Create Issues
 -------------
 
@@ -21,10 +30,10 @@ Make changes (create pull requests)
   `rendered page <https://docs.typo3.org/typo3cms/HowToDocument/Index.html>`__,
   just click on "Edit me on GitHub".
 * Step-by-step walkthrough of making a change by `Editing Directly on GitHub
-  <https://docs.typo3.org/typo3cms/HowToDocument/WritingDocsOfficial/Index.html>`__ 
+  <https://docs.typo3.org/typo3cms/HowToDocument/WritingDocsOfficial/Index.html>`__
   (this requires only a browser)
 * Step-by-step walkthrough of `Local Editing and Rendering with Docker
-  <https://docs.typo3.org/typo3cms/HowToDocument/WritingDocsOfficial/LocalEditing.html>`__ 
+  <https://docs.typo3.org/typo3cms/HowToDocument/WritingDocsOfficial/LocalEditing.html>`__
   (this requires knowledge of Git and Docker)
 
 * See `Writing Documentation <https://docs.typo3.org/typo3cms/HowToDocument/>`__ for further
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+.PHONY: help
+help: ## Displays this list of targets with descriptions
+	@echo "The following commands are available:\n"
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,12 @@
+.PHONY: help
+help: ## Displays this list of targets with descriptions
+	@echo "You prepend/append the argument 'ENV=(local|docker)' to each target. This specifies,"
+	@echo "whether to execute the target within your local environment, or docker (default).\n"
```
garvinhicking, line 4: This doesn't apply to this makefile, the env params are only evaluated in render-guides, the docker container doesn't utilize it.

garvinhicking (changes requested):

garvinhicking:
Generally I like the idea to add that to each repository that uses the new rendering. One could also add a global command individually, so any project could be rendered, but the convenience in this one as a consistent command is nice!

Makefile
linawolf: Ack

garvinhicking (approved):

CONTRIBUTING.rst
```diff
@@ -5,6 +5,15 @@ Contributing
 Information About Contributing to This Manual
 =============================================
 
+Local rendering
+---------------
+
+You can render this file locally if you have `Docker <https://www.docker.com/>`
```
brotkrueml, line 11: Which file? The contributing?

CONTRIBUTING.rst
```diff
@@ -5,6 +5,15 @@ Contributing
...
+and `make <https://www.gnu.org/software/make/>` installed on your local
+machine::
+
+    make docs
```
brotkrueml, line 15: And what is the alternative for WIndows users?

brotkrueml (commented):
I hope, we don't want to add a Makefile to all manuals?

CONTRIBUTING.rst
linawolf: WSL2.

linawolf:
> I hope, we don't want to add a Makefile to all manuals?

That would be my plan for the official manuals. And I would like to supply a snippet so ppl can add it to their extensions

CONTRIBUTING.rst
linawolf, line 11:
```suggestion
You can render this manual locally if you have `Docker <https://www.docker.com/>`
```

linawolf: changed it to manual

CONTRIBUTING.rst
brotkrueml: So we force the users to use Linux/Mac.

What about providing a `render.sh` and a `render.bat` file?

brotkrueml:
> > I hope, we don't want to add a Makefile to all manuals?
> 
> That would be my plan for the official manuals. And I would like to supply a snippet so ppl can add it to their extensions

Then we definitly need to provide a solution for Windows users.

linawolf:
I am using windows, this runs fine on WSL2 and is preinstalled. I don't know of any windows user that develops TYPO3 but does not use WSL2. And this script is not mandatory, just a convenient shortcut

brotkrueml:
Okay, then go ahead, you have already an approval.

brotkrueml:
And: Then we should document the way that works every time.

brotkrueml:
And to copy/paste: I am using Ctrl + R for searching in the zsh/bash history. It is faster than typing render.sh manually ;-)

linawolf:
For those who prefere the Docker command it still works as before...

CONTRIBUTING.rst
linawolf: We force no-one. they can use the Docker command as before?

brotkrueml:
... which is not documented here.

garvinhicking:
We could theoretically add a composer command that executes docker for the official doc repositories. That would then not have a Make/wsl2/cygwin dependency.

We could also have both of course, but I feel render-guides overdid this already.

Providing a rendering shortcut that is
- used in all official repos
- easy to run
- hides potential complexity in those long "docker run" commands
- documented
- not a requirement
- not hindering alternate rendering ways (local, manual)
- easy to maintain with low overhead

is an improvement to me? I was more used to composer scripts until I joined the project, and I was under the impression "make" was a decission made already.

brotkrueml:
IMHO Composer scripts should not be used for calling some binary stuff unrelated to PHP itself. There are other ways to do so. If the command is the same for Windows/Linux/Macos, then put it into a file called "render-docs" and that's it.

Just my 2 cents.

garvinhicking:
We'd have to add render-docs.sh and render-docs.bat then, I think that would be even less elegant than a Makefile or composer script. And clutter some directory if we add more calls to be executed (guides.xml conversion, guides.xml modification, xml linting...). I think we need a central place for it, so either Makefile (as now) or composer. Everything else would have more dependencies (ie "just" or other command runners).

PR #373, opened 2024-01-26
[TASK] Move default language to guides.xml

linawolf:
The textroles are already defined in the render-guides. the Includes.rst.txt is only needed now to display central messages.

I will include a central message about the scope of this manual (render-guides only) in a follow-up.


Documentation/guides.xml
```diff
@@ -1,5 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<guides xmlns="https://www.phpdoc.org/guides" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd" links-are-relative="true">
+<guides xmlns="https://www.phpdoc.org/guides"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="https://www.phpdoc.org/guides ../vendor/phpdocumentor/guides-cli/resources/schema/guides.xsd"
+        links-are-relative="true"
+        default-code-language="php"
```
brotkrueml, line 6: Hmm, I am not sure, if you should use this. This means we don't have to set the language to code blocks anymore when it is PHP (implicit), but AFAR we want to be explicit?

linawolf: We had the same functionality in the includes.rst.txt until now. It also applies to code blocks initialized with ::. We discourage them, however they are still used in places.

brotkrueml (approved):

PR #399, opened 2024-02-20
use correct article for the word attention

PKuhlmay:
For the word _attention_, the article "an" is used, instead of "a", because _attention_ starts with a vowel.


brotkrueml (approved):

brotkrueml:
@PKuhlmay Good catch! Thanks for the fix.

PR #402, opened 2024-03-07
[TASK] Add note about podman and optimize docker run commands

garvinhicking:
* Add a note about "podman" being interchangeable with "docker"
* Drive-by: Fixes some typos and indentation


Documentation/WritingDocForExtension/Migrate.rst
```diff
@@ -32,7 +32,8 @@ a migration tool. This tool can be used to automatically create a
 :file:`Documentation/guides.xml` from the information contained in your
 :file:`Documentation/Settings.cfg`.
 
-`Docker <https://docs.docker.com/install/>`__ needs to be installed on your
+`Docker <https://docs.docker.com/install/>`__ (or a drop-in replacement like
+`Podman <https://podman.io>`__) needs to be installed on your
```
brotkrueml, line 36: This may imply for the unexperienced user that the following command may work. Here, tabs will work.

garvinhicking: I'm sorry, I don't understand? How would you phrase/change it exactly?

It's just about saying "if you installed podman and not [only] docker, you're fine"?

brotkrueml: "Docker ... or ... Podman ... needs to be installed on your OS ... to work:

Code:  docker run ...

This implies for the unexperienced user: I install (only) Podman and can run the docker command as given in the code example? At least I read it that way.

garvinhicking: Yes

brotkrueml: So why no tabs? We do this also for CLI commands (Composer-based/Legacy).

garvinhicking: I'll try to dig into this see how it looks!👍

brotkrueml: Example code: https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/_sources/ApiOverview/CommandControllers/Index.rst.txt

garvinhicking: Was part of via #403 , resolving.

brotkrueml (approved):

PR #407, opened 2024-04-16
[TASK] Improve wording and add some additions to migration

Alagts:


Documentation/Migration/Index.rst
```diff
@@ -14,21 +14,21 @@ Migration: From Sphinx to PHP-based rendering
...
 
-The main difference that concerns you is that the PHP-based rendering
-requires a file called :file:`Documentation/guides.xml` for configuration
-while in Sphinx rendering a file called :file:`Documentation/Settings.cfg` was
-used. In the transition period we detect if a file called
-:file:`Documentation/guides.xml` is present and then switch to the
+The main difference compared to the Sphinx rendering is that the PHP-based rendering
+requires a file called :file:`Documentation/guides.xml` for configuration.
+The Sphinx rendering required a file called :file:`Documentation/Settings.cfg`.
+In the transition period the process detects if a file called
```
brotkrueml, line 20: Actually, I don't like the word "process" here. It sounds like you can run the guides rendering locally also for Settings.cfg, which is not the case. It is only related to the official rendering with our GitHub action.

Documentation/Migration/Index.rst
```diff
@@ -59,13 +59,25 @@ operating system for the tool to work:
...
+    ..  code-block:: text
+        :caption: Output of the command
+
+        Note: Some of your settings could not be converted:
+            * html_theme_options
+            * project_discussions
+            * use_opensearch
+
+everything went well. They can be ignored since these files are usually files that
+not need to be converted. You now can safely delete :file:`Settings.cfg`. Also delete
```
brotkrueml, line 74: "You now can safely ..." sounds strange.

Documentation/Migration/Index.rst
```diff
@@ -78,6 +90,20 @@ PHP-based rendering tool:
 
 ..  rst-class:: bignums
 
+#.  Manual modifications of the :file:`guides.yml`
+
+    You have to manually change the following: in the extension tag add the attribute
```
brotkrueml, line 95:
```suggestion
    You have to manually change the following: in the :xml:`<extension>` tag add the attribute
```

Documentation/Migration/Index.rst
```diff
@@ -78,6 +90,20 @@ PHP-based rendering tool:
...
+
+        ..  code-block:: text
+            :caption: Changes in your :file:`guides.yml`
+
+            interlink-shortcode="vendor/ext_name"
+
+    This you can find in your composer.json under "name". E.g. :json:`friendsoftypo3/jumpurl`.
```
brotkrueml, line 102: "This you can find ...." sounds strange.

brotkrueml, line 102:
```suggestion
    This you can find in your :file:`composer.json` under "name". E.g. :json:`friendsoftypo3/jumpurl`.
```

Documentation/Migration/Index.rst
```diff
@@ -78,6 +90,20 @@ PHP-based rendering tool:
...
+    We recommend to reformulate the code using the mac short cut `cmd + option/alt + l` and to
```
brotkrueml, line 103: What about Windows/Linux users? No Max shortcut? Additionally, use the :kbd: role.

Additional "reformulare" sounds strange, better "reformat"

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+Interlink inventory not found: HTTP/2 404
+"""""""""""""""""""""""""""""""""""""""""
+
+.. code-block:: text
+
+    [2024-03-13T12:22:50.661532+00:00] app.WARNING: Interlink inventory not found: HTTP/2 404
+    returned for "https://docs.typo3.org/m/typo3/book-extbasefluid/11.5/en-us/objects.inv.json". [] []
+
+Here you see that the link `https://docs.typo3.org/m/typo3/book-extbasefluid/11.5/en-us` is not found (404 page).
+We can now check via google if there is another link to book Extbasefluid. We found this site `https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/`
```
brotkrueml, line 227:
```suggestion
We can now check via Google if there is another link to book Extbasefluid. We found this site `https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/`
```

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+We can find the hint: `This manual is no longer being maintained for TYPO3 versions 11.5 and above.`. This tells us that the developers stopped to update the
```
brotkrueml, line 228:
```suggestion
We can find the hint: `This manual is no longer being maintained for TYPO3 versions 11.5 and above.`. This tells us that the Documentation Team abandoned this
```

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+documentation. We can for example link to the last existing version. Which is 10.4. To do this we have to change the :file:`guides.xml`. Search for the
```
brotkrueml, line 229:
```suggestion
manual. We can, for example, link to the last existing version. Which is 10.4. To do this we have to change the :file:`guides.xml`. Search for the
```

Use the outdated 10.4 is not recommended. Instead, the content should be searched in the other manuals, mainly TYPO3 Explained.

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+Inventory link with key ... not found
+"""""""""""""""""""""""""""""""""""""
+
+.. code-block:: text
+
+    [2024-03-13T12:26:40.940930+00:00] app.WARNING: Inventory link with key "h2document:rest-common-pitfalls"
+    (rest-common-pitfalls) not found.  {"rst-file":"GeneratedExtension/Index","type":"ref","targetRef
+
+We see already that we have to go to file :file:`GeneratedExtension/Index` in the directory "Documentation".
+In there we have to delete the line which contains
```
brotkrueml, line 251: I don't think deleting is always the best case. Better would be to look for the content in the same manual, but maybe another anchor.

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. code-block:: text
+
+    [2024-03-25T13:26:11.600367+00:00] app.WARNING: Nested PHP domain components
+    (php:class, php:interface, php:enum etc) are not supported.
+    Found php:\Vendor\MyExtension\Interfaces\RequireJsModuleInterface inside
+    \Vendor\MyExtension\Interfaces\AnotherImportantInterface {"rst-file":"Developer/Index.rst"} []
+
+The file :file:`Index.rst` in Documentation/Developer has a wrong indentation. A class must not belong to another class.
```
brotkrueml, line 269:
```suggestion
The file :file:`Index.rst` in :file:`Documentation/Developer/` has a wrong indentation. A class must not belong to another class.
```

We always append a slash at the end of folders to distinguish between folders and files.

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+Additionally regarding the name it has to be interfaces and not classes.
+`..  php:class:: ExampleInterface` has to be changed to `..  php:interface:: ExampleInterface`.
+
+
+..  _migrate-reference-could-not-be-resolved:
+
+Reference sitehandling-addinglanguages could not be resolved in LocalizedContent/Index {"rst-file":"LocalizedContent/Index"} []
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+The next step is to visit the old site, e.g. `https://docs.typo3.org/m/typo3/guide-frontendlocalization/main/en-us/LocalizedContent/Index.html`.
```
brotkrueml, line 292: Why **old** site?

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+There we can search for `sitehandling-addinglanguages` in the rst code by clicking the button "`</>View Source`".
+Finding this:
+
+.. code-block:: rst
+
+    ..  tip::
+    For more information on how to add languages and configure their
```
brotkrueml, line 299: Wrong indentation.

Documentation/Migration/Index.rst
linawolf, line 20:
```suggestion
In the transition period the GitHub action of the official rendering process detects if a file called
```

linawolf: How about my suggestions or what would you suggest?

Documentation/Migration/Index.rst
linawolf, line 74:
```suggestion
not need to be converted. You can now delete :file:`Settings.cfg`. If you ever should need it again it is still in the git history. Also delete
```

Documentation/Migration/Index.rst
linawolf: When you are in a process of migrating the old documentation this is what you would do. And then recommend to replace the links in follow up work?

Documentation/Migration/Index.rst
linawolf: In this case the whole section was removed. It depends on the use case what the best solution looks like here...

Documentation/Migration/Index.rst
linawolf: Because you can see if the refernce was still working in the old rendering?

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,121 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+    behaviour in the site configuration, see
+    :ref:`Adding Languages <sitehandling-addinglanguages>`.
```
linawolf, line 301:
```suggestion
    ..  tip::
        For more information on how to add languages and configure their
        behaviour in the site configuration, see
        :ref:`Adding Languages <sitehandling-addinglanguages>`.
```

Documentation/Migration/Index.rst
Alagts: @brotkrueml Suggest a nice sentence then I commit it.

Documentation/Migration/Index.rst
Alagts: When I remember correctly there was no other option than linking to the old documentation. I am not sure if linking to another article would be better here. This decision is best addressed to the owner of the extension builder.

Documentation/Migration/Index.rst
Alagts: By "old" I meant the old links when using the Sphinx Rendering. As "new" I would refer to the PHP-based rendering.

Documentation/Migration/Index.rst
brotkrueml: That is what I mean: You present a recipe, users follow it (maybe without thinking about it), which is maybe not what would be the best case.

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,134 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+We conclude: In general it depends on the case itself what the best solution is.
+
+..  _migrate-nested-php-domain-components-not-supported:
+
+Nested PHP domain components (php:class, php:interface, php:enum etc) are not supported
+"""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+
+.. code-block:: text
+
+    [2024-03-25T13:26:11.600367+00:00] app.WARNING: Nested PHP domain components
```
brotkrueml, line 268: Actually, I don't understand the whole text of the warning. But this is another story.

But please use "MyVendor" instead of "Vendor" in the examples.

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,134 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+    (php:class, php:interface, php:enum etc) are not supported.
+    Found php:\Vendor\MyExtension\Interfaces\RequireJsModuleInterface inside
+    \Vendor\MyExtension\Interfaces\AnotherImportantInterface {"rst-file":"Developer/Index.rst"} []
+
+The file :file:`Index.rst` in :file:`Documentation/Developer/` has a wrong indentation. A class must not belong to another class.
+Here is the wrong rst code.
+
+.. code-block:: rst
```
brotkrueml, line 276: What is wrong in this code? I see no wrong indentation?

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,134 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+
+    Used for ...
+
+    ..  php:class:: RequireJsModuleInterface
+
+    Widgets implementing this interface will add the provided RequireJS modules.
+    Those modules will be loaded in dashboard view if the widget is added at least once.
+
+Additionally regarding the name it has to be interfaces and not classes.
+`..  php:class:: ExampleInterface` has to be changed to `..  php:interface:: ExampleInterface`.
```
brotkrueml, line 288:
```suggestion
:rst:`..  php:class:: ExampleInterface` has to be changed to :rst:`..  php:interface:: ExampleInterface`.
```

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,134 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+
+.. image:: /Images/get_link.png
+    :class: with-shadow
+    :width: 600px
+
+.. code-block:: rst
+
+    :ref:`Adding Languages <t3coreapi:sitehandling-addingLanguages>`
+
+You have to replace the correct link in e.g. :file:`Documentation/LocalizedContent/Index.rst` to
```
brotkrueml, line 321:
```suggestion
You have to replace the correct link in, for example, :file:`Documentation/LocalizedContent/Index.rst` to
```

Documentation/Migration/Index.rst
```diff
@@ -174,6 +200,134 @@ of typing a long :bash:`docker run...` or :bash:`podman run...` command:
...
+fix the error. Note: Since we already migrated the Frontend Localization Guide
+(`https://docs.typo3.org/m/typo3/guide-frontendlocalization/main/en-us/LocalizedContent/Index.html`)
+to use the PHP based rendering you cannot find the state that we have shown above anymore.
```
brotkrueml, line 324:
```suggestion
to use the PHP-based rendering you cannot find the state that we have shown above anymore.
```

brotkrueml (changes requested):
Regarding the warning examples: I suggest to add a paragraph, like

    Problem: anchor not found

or

    Problem: manual in version not found

directly after the heading. I have to read the whole section to get what is wrong. This short introduction eases the understanding of the problem.

Documentation/Migration/Index.rst
brotkrueml: But how can a user visit an "old site" when it is already rendered with guides?

As you see, I haven't understand it, so maybe our readers also don't.

Documentation/Migration/Index.rst
Alagts: I do not think that it is a good practice to manipulate warnings. I mean it came up like this, why I should change the "Vendor" name to "MyVendor?". In the other examples I agree but here I disagree.

Documentation/Migration/Index.rst
Alagts: Me as a user wants to have at least a small hint what can be the correct behavior than having nothing at hand. I feel not professional enough to improve this article to satisfy your thoughts. So maybe another person can overtake this task.

Documentation/Migration/Index.rst
brotkrueml: It is an example for an error message. We do not provide the original source, do we? So this can be adjusted. But I don't insist for change, but like it for consistency reasons.

Documentation/Migration/Index.rst
linawolf: @brotkrueml can you try to improve this section in a follow up please. This is beyond the scope of this PR

Documentation/Migration/Index.rst
linawolf: @brotkrueml can you try to improve this section in a follow up please. This is beyond the scope of this PR

Documentation/Migration/Index.rst
linawolf: @brotkrueml can you try to improve this section in a follow up please. This is beyond the scope of this PR

Documentation/Migration/Index.rst
linawolf: Because "A class must not belong to another class." See sentence above

linawolf:
@brotkrueml please let us get this merged. You are welcome to further improve this section in a follow up

Documentation/Migration/Index.rst
brotkrueml: Personally, I would remove this section from here as this has nothing to do with the migration. Maybe move it later to a "Troubleshooting" section which is independent from the migration (and can also be linked from here). We should distinct between warnings appearing because of the migration and warning appearing for other reasons.

I may assume that people don't look into the migration chapter when warning appear in a year or two, but maybe into a troubleshooting chapter.

But fine for me for now, not blocking it.

brotkrueml:
> @brotkrueml please let us get this merged. You are welcome to further improve this section in a follow up

I am not blocking it. Feel free to approve and merge, I won't approve it for now.

linawolf (approved):
Thanks a lot for these great changes

PR #410, opened 2024-04-17
[TASK] Remove Hint

linawolf:
The same information is already contained in the general disclaimer

### DIFF
--- a/Documentation/WritingDocForExtension/Index.rst
+++ b/Documentation/WritingDocForExtension/Index.rst
@@ -10,10 +10,6 @@
 How to document an extension
 ============================
 
-..  hint::
-    If your extension does not have a file :file:`EXT:my_extension/Documentation/guides.xml`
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
Where can I find the "general container"?

linawolf:
Index.rst.txt

linawolf:
Or just look at it rendered: https://docs.typo3.org/m/typo3/docs-how-to-document/main/en-us/WritingDocForExtension/Index.html

brotkrueml:
Hmm, I suggest to leave it as is: We link directly from "docs.typo3.org > Developing and using extensions > Document an extension" to this page. So this information might get lost as users won't mostly navigation to the start page from this page.

linawolf:
The Disclaimer is displayed on top of EVERY page

Alagts (approved):

PR #417, opened 2024-04-21
[TASK] Delete page special styles

linawolf:
We do not want to encourage ppl to use classes


brotkrueml (commented):
Then we should also adjust our manuals to avoid those classes (figure).

PR #425, opened 2024-05-17
[FEATURE] Document how to Regenerate existing code snippets

linawolf:


Documentation/Maintainers/Codesnippets.rst
```diff
@@ -0,0 +1,60 @@
...
+
+        ..  code-block:: bash
+
+            make install
+
+    ..  tab:: composer
+
+        ..  code-block:: bash
+
+            # in repositories where the composer.lock is not in version control
```
garvinhicking, line 39: Wouldn't "composer install" be the way here? 

Where composer.lock exists, this would be respected and install locked versions.
Where composer.lock does not exist, the install would create it.

Using "composer update" would override composer.lock versions in this case, which could be problematic?!

Documentation/Maintainers/Codesnippets.rst
```diff
@@ -0,0 +1,60 @@
...
+            make codesnippets
+
+    ..  tab:: TYPO3 console command
+
+        ..  code-block:: bash
+
+            # Paths depends on where the vendor binaries are installed
+            # And where the configuration file codesnippets.php is located
+
+            vendor/bin/typo3 codesnippet:create Documentation/
```
garvinhicking, line 60: Just so I understand: Every repo that has code snippets will have a linked TYPO3 installation (via composer dependency) and sets up a command that reacts on "codesnippet:create"? Or is that action provided through another repository (codesnippets?) and pulled as a dependency?

Clearing that up would help to understand how the whole process works...

Documentation/Maintainers/Codesnippets.rst
linawolf: Actually when you don't have the composer.json in VCS you want to always do composer update. As you still have a local composer.lock. Otherwise you will not get the newest version like someone who makes a fresh install would have

Documentation/Maintainers/Codesnippets.rst
linawolf: Currently the codesnippet tool is a TYPO3 extension. However this would belong in other sections.

Documentation/Maintainers/Codesnippets.rst
mavolkmer: The HowTo-text starts with cloning a repo - how should a local composer.lock exist if there is none in repo?

linawolf: Its a common mistake that we make if we already cloned the repository and installed it before. To do composer install and not composer update and then you will have outdated codesnippets.

mavolkmer: ok, I understand - but then I would suggest to mention that directly

and what about repositories where composer.lock is in VCS? As you addressed the case that it is not in VCS we should also mention the other case, don´t we?

```suggestion
            # in repositories with composer.lock in version control
            composer install
            
            # in repositories where the composer.lock is not in version control
            # to prevent outdated packages due to a local composer.lock from earlier install-runs 
```

garvinhicking (approved):
Thanks for the clarification! :)

PR #427, opened 2024-06-10
[TASK] Add reference to typo3-documentation-rendersync project

garvinhicking:
My "pet project" typo3-documentation-browsersync is in a state where I'd like to collect some feedback, also to see if in the end this project might be movable to the TYPO3-Documentation repository.

For that I've created a small documentation on how to use it, hopefully also a base to present it for the T3DD24 in the scope of Linas talk presenting render-guides.

(This also drive-by-fixes an issue with a missing literalainclude syntax for _Makefile that removes a warning from rendering. Another issue is still open that causes the rendering to fail with a duplicate RST header, which is beyond the scope of this PR)

### DIFF
--- a/Documentation/RenderingDocs/Index.rst
+++ b/Documentation/RenderingDocs/Index.rst
@@ -68,6 +68,8 @@ Make sure that `Docker <https://www.docker.com/>`__ is installed on your system.
             docker run --rm --pull always -v ${PWD}:/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation
             start "Documentation-GENERATED-temp/Index.html"
 
+..  _rendering-wysiwyg:
+
 Rendering with more WYSIWYG-feeling (automatic re-rendering)
 ------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


linawolf:
@garvinhicking please rebase then the rendering warnings should be resolved

Documentation/RenderingDocs/Index.rst
```diff
@@ -68,6 +68,89 @@ Make sure that `Docker <https://www.docker.com/>`__ is installed on your system.
             docker run --rm --pull always -v ${PWD}:/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation
             start "Documentation-GENERATED-temp/Index.html"
 
+Rendering with more WYSIWYG-feeling (automatic re-rendering)
```
linawolf, line 71:
```suggestion
..  _rendering-wysiwyg:

Rendering with more WYSIWYG-feeling (automatic re-rendering)
```

Please introduce a unqie anchor if you introduce a headline

Documentation/RenderingDocs/Index.rst
```diff
@@ -68,6 +68,89 @@ Make sure that `Docker <https://www.docker.com/>`__ is installed on your system.
...
+------------------------------------------------------------
+
+Often, especially in the later stages of creating documentation, you
```
linawolf, line 74: I would start with a sentence about what it does. Otherwise you might loose readers in the first paragraph

garvinhicking: I added such a sentence, but it's not really my style ;) I always first want to establish the groundrules so for me I'd first want to postulate the problem/setup and then work from there on how to improve it.

But maybe this works now?

linawolf: Better. We are writing for the internet here and ppl are volatile. Studys show they decide within the first few seconds if they keep reading or not.

linawolf (approved):

PR #431, opened 2024-06-13
[TASK] Enhance glossary by specific product names

garvinhicking:
TYPO3 has its own rules, so let's respect upper/lowercasing from others as well :-)

Follow-up to https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/4509#discussion_r1632168380

### DIFF
--- a/Documentation/GeneralConventions/Glossary.rst
+++ b/Documentation/GeneralConventions/Glossary.rst
@@ -125,7 +125,7 @@ P
 `PHP CS Fixer`,
 `PHP_CodeSniffer`,
 `PNG`,
-`PostgreSQL`,
+`PostgreSQL`
 
 Q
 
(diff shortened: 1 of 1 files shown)


Documentation/GeneralConventions/Glossary.rst
```diff
@@ -119,6 +119,11 @@ O
 P
 
 `page TSconfig`,
+`PhpStorm`,
+`PHPStan`,
+`PHPUnit`,
+`PHP CS Fixer`,
+`PHP_CodeSniffer`,
```
froemken, line 126: With underscore? OK...in about-section of their GitHub repo it was also written that way. Ugly. Brrr. Approved.

froemken (approved):

Documentation/GeneralConventions/Glossary.rst
garvinhicking: Yes in their Wiki and everywhere they utilize this spelling. Well... some people also thing TYPO3 is odd :D

PR #437, opened 2024-07-05
[TASK] Add missing attribute `project_discussions`

ErHaWeb:
Add attribute `project_discussions`, as a corresponding GitHub page exists in the context of the news extension from the example.


froemken (approved):
Looks good to me. But someone else should approve, as discussions are not activated for all GitHub repos by default.

ErHaWeb:
That's true, but as an extension author you might want to think about whether a discussion page in your own repository makes sense. If you neither have nor want such a page, you can always remove the attribute. For me, this example is more about showing the possibilities.

garvinhicking:
Thanks, I agree for the example we should show the whole bouqet. Good idea!

PR #459, opened 2024-10-15
Fix branch for docker-render-documentation

pschriner:
The docker-render-documentation repository is working on the master branch, not main.


pschriner:
tbh this should *not* be merged. Rather it should be adapted to the new php based rendering

PR #460, opened 2024-11-04
[DOC] correct wrongly used filename

kaystrobach:
use guides.xml instead of yaml


garvinhicking (approved):

garvinhicking:
CI run error caused by unrelated change, merging this despite of it. Thanks!

PR #477, opened 2024-12-17
[FEATURE] Document new init command

linawolf:


Documentation/Reference/GuidesXml.rst
```diff
@@ -9,19 +9,33 @@ Configuration of the rendering - guides.xml
...
-
 ..  literalinclude:: _guides-simple.xml
     :caption: Documentation/guides.xml
 
+..  contents:: Table of Contents:
+    :local:
+
+..  _guides-xml-init:
+
+Init new guides.xml
```
jaapio, line 20: I would reword this, what is guides.xml? 

`initialize new documentation project`

linawolf: We are on the page, which describes what the guides.xml does here. Therefore I think in this context it is ok

PR #495, opened 2025-06-08
[TASK] Create chapter for inline markup

linawolf:
Move the content away from the mix-up with
the inline-code roles. Link to related elements like links and inline code.


sarahmccarthy123 (approved):
Looks good :-)

PR #509, opened 2026-02-09
crosslink Inline code chapters

franzholz:
It is not evident starting from chapter "Inline code with or without overlays" that a more detailed chapter "Inline code" already exists.

### DIFF
--- a/Documentation/Reference/ReStructuredText/Code/InlineCode.rst
+++ b/Documentation/Reference/ReStructuredText/Code/InlineCode.rst
@@ -33,7 +33,7 @@ By convention single backticks are used unless double are needed.
 Code roles with language information and overlay
 ================================================
 
-You can also use text roles with one of the predefined languages to display more
+You can also use `text roles <https://docs.typo3.org/permalink/h2document:text-roles>`_ with one of the predefined languages to display more
 information to the user. For the most common languages, automatic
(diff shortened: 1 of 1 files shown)


Documentation/Reference/ReStructuredText/Code/InlineCode.rst
```diff
@@ -33,7 +33,7 @@ By convention single backticks are used unless double are needed.
 Code roles with language information and overlay
 ================================================
 
-You can also use text roles with one of the predefined languages to display more
+You can also use `text roles <https://docs.typo3.org/permalink/h2document:text-roles>`_ with one of the predefined languages to display more
```
DavidBruchmann: I'm not too deep into it, but shouldn't it be a different link notation?
At least without the domain, and as it's a permalink **perhaps** a completely different internal notation should be used.

DavidBruchmann: that's a surprising proposition.

DavidBruchmann:
```suggestion
You can also use :ref:`text roles <h2document:text-roles>`_ with one of the predefined languages to display more
```

linawolf (approved):

PR #534, opened 2026-08-11
[TASK] Add shared AGENTS.md/CLAUDE.md/STYLEGUIDE.md, reference from CONTRIBUTING

linawolf:
Track AGENTS.md and STYLEGUIDE.md, previously local-only files, as
the shared home for writing/workflow conventions across TYPO3
documentation repos. Generalized their content from an earlier
CoreApi-specific draft: kept the rules that apply everywhere, made
repo-specific assumptions (tooling, branch structure, example
anchors, version numbers) conditional or current instead.

CLAUDE.md is a short pointer to AGENTS.md, kept only because Claude
Code specifically looks for that filename; AGENTS.md itself is
tool-agnostic so other assistants (and other repos) can use it too.

CONTRIBUTING.rst now links all three under a new "Writing style"
section, explaining that other repos should reference these rather
than duplicate them.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: Lina Wolf


CybotTM:
Did you mix up three different concerns in this PR? Repeating the project's content for an LLM? Providing LLM instructions for the repository? Providing generic TYPO3 documentation instructions?

Besides this, you explain some agent behaviours that belong in the agent's own harness, not in the project's `AGENTS.md`.

`AGENTS.md` does not need to repeat existing information. It just has to point to the correct places for contribution guidelines, its own documentation as the documentation guideline, and so on.

So `AGENTS.md` only needs to point to `CONTRIBUTING.rst` for contribution guidelines, `Documentation/` for how to document, and so on.

I would also avoid explaining how to use Git.
The detailed mechanics should be enforced by GitHub permissions, the harness, not by rules.

You should start with a small `AGENTS.md` and add only things that solve real issues that appeared during agent sessions.

And you should evaluate them: ask your agent to solve a task with and without `AGENTS.md`, or with and without a specific change to `AGENTS.md`. A change should only be applied if it proves to improve the agent's work.

linawolf:
Ok so it seems like LLMs can read our rst files instead of relying on their own STYLEGUIDE.md instead we will build in improvements directly into the rst files

sarahmccarthy123 (approved):
Looks good :-)

linawolf:
@CybotTM I only added the Agents specific things now and will integrate my findings from the styleguide into the rst files

CybotTM:
yes, looks good to me.

PR #540, opened 2026-08-14
[TASK] Warn that --output is resolved inside the container

CybotTM:
The `--output` option of the render container is resolved inside the container: pointing it at a host path outside the mounted volume writes the files into the container file system, which `--rm` then discards — while the command still reports success. This adds a note to the shared local-rendering include so every page showing the command carries the warning.


linawolf (commented):
The example command does not show the usage of the output parameter and usually it is not used by ppl just trying out the rendering. So I am wondering if this note block is really needed, expoaccialy since this include is included in so many places throughout the manual

CybotTM:
Moved in f855eb2. You are right that it does not belong there: `_LocalRendering.rst.txt` is included in four pages and its example never passes `--output`. The only page in this manual that does is the Markdown-to-reST howto, so the note now sits next to those commands and is down to the point that matters — the path is resolved inside the container and has to stay below the mounted volume, and the relative `Documentation-Migrated` used there already satisfies that. The shared include is back to what it was. `make test-docs` renders clean.

linawolf (approved):

PR #542, opened 2026-08-14
[TASK] Comments: document that todo notes are dropped on render

CybotTM:
Adds a short section that todo notes must not remain in published documentation — resolve them before merging or demote them to author-only comments.

### DIFF
--- a/Documentation/Reference/ReStructuredText/Content/Comments.rst
+++ b/Documentation/Reference/ReStructuredText/Content/Comments.rst
@@ -52,3 +52,12 @@ Example:
 
     *   three
 
+..  _writing-rest-comments-todo:
+
+Open tasks (todo)
+-----------------
+
+The TYPO3 documentation renderer silently drops a :rst:`..  todo::`
+directive — its content never reaches the rendered page. Do not rely on it
+as a visible marker: resolve open tasks before merging, or write them as
+plain comments (dot-dot-blank) so they stay clearly author-only.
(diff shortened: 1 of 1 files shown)


linawolf (commented):
I'd drop the "or write them as plain comments" alternative. I implemented                                                                                                                                                      
 `..  todo::` myself — not rendering it is by design, not a quirk. It's meant                                                                                                                                                   
 to be an internal todo comment, and a better one than a bare `..` comment,                                                                                                                                                     
 since `todo::` is a distinct, greppable string (`..` alone also starts every                                                                                                                                                   
 directive and anchor).                                                                                                                                                                                                         
                                                                                                                                                                                                                                
 I'd flip the guidance: keep using `..  todo::` for internal notes, just warn                                                                                                                                                   
 that it gives no visual reminder in the rendered output — grep for `todo::`                                                                                                                                                    
 before merging rather than relying on noticing it in a preview.

CybotTM:
Flipped in baf5a8d, thank you — you are right that the plain-comment alternative was the wrong advice. The section now says to keep using `..  todo::` for internal notes, gives your reason for preferring it over a bare `..` comment (a distinct, greppable string, where `..` also starts every directive and anchor), and warns only about the consequence: no visual reminder in the rendered page, so search for `todo::` before merging. `make test-docs` renders clean.

linawolf (approved):

PR #544, opened 2026-08-14
[TASK] Versions directives: reference released versions only

CybotTM:
Documents that version directives must reference released TYPO3 versions only — plans change, published versionadded directives stay.

### DIFF
--- a/Documentation/Reference/ReStructuredText/Content/Versions.rst
+++ b/Documentation/Reference/ReStructuredText/Content/Versions.rst
@@ -12,6 +12,12 @@ We have 3 directives to mark versions in the manuals.
 The incentive is that information which is important for migration can be marked
 and semi-automatically be removed after one or two versions.
 
+..  note::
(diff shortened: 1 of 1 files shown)


linawolf (commented):
That note is actually not correct. While the Core Team works on TYPO3 15.0 we already document it in the main version using versionadded:: 15.0 etc. We are trying to have the documentation ready when a release is beeing made and that means we are working on it before the version has actually been released.

CybotTM:
Closing this — the note is simply wrong, thank you for catching it. I assumed version directives should trail the releases, and that is not how the Core Team works: documenting TYPO3 15.0 with `versionadded:: 15.0` while 15.0 is being built is the point, so the documentation is ready when the release is. A note telling authors to reference released versions only would have contradicted the established practice and the manual's own history.

Nothing here is worth salvaging under a different wording, so I would rather withdraw it than rewrite it into something narrower.

PR #545, opened 2026-08-14
[TASK] guides.xml: inventory slash, extension class, theme attr

CybotTM:
Three small additions to the guides.xml reference, all from observed failures: inventory URLs should end with a trailing slash; a missing extension class silently falls back to the default phpDocumentor layout (no error raised); the schema-valid theme attribute is not needed (the theme comes from the extension class) while a theme child element aborts rendering with 'Invalid type for path guides.theme'.

### DIFF
--- a/Documentation/Reference/GuidesXml.rst
+++ b/Documentation/Reference/GuidesXml.rst
@@ -50,6 +50,13 @@ And follow the interactive question.
 
         The following settings can be relevant for TYPO3 themed documentation:
 
+        ..  note::
+            The schema accepts a :xml:`theme` attribute on :xml:`<guides>`,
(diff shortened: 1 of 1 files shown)


Documentation/Reference/GuidesXml.rst
```diff
@@ -242,6 +252,12 @@ And follow the interactive question.
         The class attribute is mandatory, it references the extension that is used
         to render the documentation with the TYPO3 documentation theme.
 
+        ..  note::
+            If the :xml:`class` attribute (or the whole :xml:`<extension>`
```
linawolf, line 256: We should require the theme to be set in the render guides.

linawolf (approved):

PR #547, opened 2026-08-14
[DOCS] literalinclude: marker-based selection hint

CybotTM:
Mentions the partial-inclusion options (`:lines:`, and the more edit-robust `:start-after:`/`:end-before:` markers) and links the underscore naming convention for include-only files.

### DIFF
--- a/Documentation/Reference/ReStructuredText/Code/Codeblocks.rst
+++ b/Documentation/Reference/ReStructuredText/Code/Codeblocks.rst
@@ -377,6 +377,12 @@ code block.
             :emphasize-lines: 5,10-13
             :linenos:
 
+The directive can display parts of a file, for example with :rst:`:lines:`
+or — more robust against later edits — with the markers :rst:`:start-after:`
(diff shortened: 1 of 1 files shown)


CybotTM:
Closing after a deeper check: the TYPO3 renderer does not implement literalinclude line selection at all — typo3-docs-theme's LiteralincludeDirective renders the whole file, and phpDocumentor/guides' option mapper knows only language/caption/emphasize-lines/linenos/name. Documenting :lines:/:start-after:/:end-before: here would describe Sphinx behaviour the toolchain does not have — the same class of problem that #490 removed for the :diff: example (render-guides#974). If line selection ever lands in render-guides, this can be revived.

linawolf:
Should we open an issue in the phpdocumentor/guides about thisß

CybotTM:
@linawolf 

tried my best:

- phpDocumentor/guides#1351
- phpDocumentor/guides#1352
- phpDocumentor/guides#1353

but I am not feeling lucky they will get through.

